PR #16, opened 2026-05-19
AI Suggestions Feature fixes and Testing set up

### LINKED ISSUE #4: CLO to PLO and CLO to Accreditation Standard Mapping
# DESCRIPTION

The system should allow program designers to generate suggested CLO to PLO and CLO to Accreditation Standard mappings using generative AI.

(More details to be added later.)

Apps247:
### Code refactor: Bootstrap 5 and test consolidation
- **Migrations:** Fixed three broken `down()` methods (table-name typo in `2025_12_17_094210`, empty rollbacks in `2025_12_15_153457` and `2025_09_23_112522`)
- **Testing:** Added `.env.testing.example`, and consolidated tests into `composer test`, so that tests now use a separate DB. This DB must be created by the user manually. If it is not found, the precheck script `scripts/check-test-db.php` that automatically runs with `composer test` hints to run the test DB creation command
- **Vite:** Removed the `Vite::useScriptTagAttributes` override that was breaking ES module loading in dev
    -  **jQuery:** The build/dev scripts copy jQuery from `node_modules` into `laravel/public/vendor` to serve it to the frontend
- **Gitignore:** Removed build artifacts from source control
- **Bootstrap 5:** Upgraded frontend code to all use bundled BootStrap `^5.3.3`
- **Dependency patches:** `axios ^0.25` to `^1.7.7`, `google/recaptcha 1.2.*` to `^1.3`, `phpoffice/phpword ^0.18.1` to `^1.3`; dropped unused `lodash`, `resolve-url-loader`, and `sass-loader`. `npm audit` and `composer audit` now report clean.
- **Docs:** Rewrote `docs/Setup.md` with embedded `.env.example`/`.env.testing.example` snippets, PostgreSQL cluster init steps (`initdb`, `pg_ctl`, role and database creation under `-U postgres`), the Vite dev/build workflow, and the `composer test` flow.

### AI Suggestions Feature

Ticks off leftover items from [PR8](https://github.com/forestryTLS/curriculum-development-tool/pull/8)

- Integrated AWS Lambda with Laravel backend
- Set up on-demand pull of SageMaker output from DynamoDB in the frontend.
- When the page loads, it attempts to fetch any AI suggestions for that PLO-CLO pairs that weren't fetched yet. It keeps polling for 10 minutes - if the user stays on the page for 10 minutes, it stops polling and shows a refresh button which restarts the polling when clicked. If the suggestions are available, they're pulled, stored in the postgreSQL database, and the page is reloaded to display the suggestions.
If SageMaker has `is_mapped: false` for a certain PLO-CLO pair, then it is mapped to the column labelled "N/A"
- Bug fixes
- Added localstack e2e tests for the lo mapping service. They don't test SageMaker itself, but do test the frontend and laravel portions of the flow. These run with the moto unit tests with `pytest -v -s`. Details on running tests found in `lo_mapping_service/tests/e2e/README.md`


laravel/app/Http/Controllers/CourseProgramController.php
```diff
@@ -184,4 +188,264 @@ public function updateAiSuggestionStatus(Request $request, $courseId, $programId
...
+            Log::info("Generating AI suggestions (MOCK MODE) for course_id=$courseId, program_id=$programId");
+
+            return response()->json([
+                'status' => 'mock',
+                'message' => 'Mock AI suggestions generated successfully',
+                'request_id' => 'mock_request_' . uniqid(),
+                'course_id' => $courseId,
+                'program_id' => $programId,
+            ]);
+        }
```
FabianLCH: I'm assuming you added this for some quick manual testing, right? If so, not super important, but we can maybe clean this up later before merging

laravel/app/Http/Controllers/CourseProgramController.php
```diff
@@ -184,4 +188,264 @@ public function updateAiSuggestionStatus(Request $request, $courseId, $programId
...
+            ], 500);
+        }
+    }
+
+    /**
+     * Query the lo_mapping_service for in-flight status of multiple (course, program) pairs.
+     * Returns ['<programId>' => bool, ...] for the given course/program pairs.
+     * Used at Step 5 render time and for pre-click race checks.
+     */
+    public static function getInFlightStatuses(int $courseId, array $programIds): array
```
FabianLCH, line 277: It looks like this function always maps a single `$courseId` to each program ID in `$programIds`. Maybe we can save ourselves some work here and send the single course ID (either in the POST data or as part of the path) and the list of program IDs to the FastAPI endpoint instead of creating the `$pairs` array

laravel/app/Http/Controllers/CourseProgramController.php
```diff
@@ -184,4 +188,264 @@ public function updateAiSuggestionStatus(Request $request, $courseId, $programId
...
+
+            $statuses = $response->json('statuses') ?? [];
+            $result = array_fill_keys($programIds, false);
+            foreach ($statuses as $entry) {
+                $pid = (int) ($entry['program_id'] ?? 0);
+                $result[$pid] = (bool) ($entry['in_flight'] ?? false);
+            }
+            return $result;
+        } catch (\Exception $e) {
+            Log::warning('getInFlightStatuses exception: ' . $e->getMessage());
```
FabianLCH: Should probably log this with `Log::error(...)`

laravel/app/Http/Controllers/CourseProgramController.php
```diff
@@ -184,4 +188,264 @@ public function updateAiSuggestionStatus(Request $request, $courseId, $programId
...
+            return array_fill_keys($programIds, false);
+        }
+    }
+
+    public function checkInFlight(Request $request, $courseId, $programId)
+    {
+        $statuses = self::getInFlightStatuses((int) $courseId, [(int) $programId]);
+        return response()->json([
+            'in_flight' => $statuses[(int) $programId] ?? false,
+        ]);
```
FabianLCH, line 311: Is all the type casting here necessary? Eloquent models should already ensure course and program IDs returned by the database are of type `int`

laravel/app/Http/Controllers/CourseProgramController.php
```diff
@@ -184,4 +188,264 @@ public function updateAiSuggestionStatus(Request $request, $courseId, $programId
...
+    }
+
+    public function checkAiResults(Request $request, $courseId, $programId)
```
FabianLCH: We should add a comment explaining this function checks for records (jobs?) awaiting processing and triggers processing if so

laravel/app/Http/Controllers/CourseProgramController.php
```diff
@@ -184,4 +188,264 @@ public function updateAiSuggestionStatus(Request $request, $courseId, $programId
...
+                        'map_scale_id'  => $abbreviationToScaleId[$label],
+                    ]);
+                    $rowsWritten++;
+                }
+            }
+
+            $courseProgram = CourseProgram::where(['course_id' => $courseId, 'program_id' => $programId])->first();
+            if ($courseProgram) {
+                $courseProgram->ai_suggestion_status = true;
+                $courseProgram->manual_map_status    = true;
```
FabianLCH: Unless I'm missing something else that depends on it, we probably don't want to modify `manual_map_status` here as this function only deals with the AI suggestions

python/services/lo_mapping_service/tests/e2e/test_ai_suggestions_e2e.py
FabianLCH, line 1: The E2E tests should definitely not be done from within the `lo_mapping_service`'s context. Also, this mixes integration tests with E2E tests.

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
```diff
@@ -130,13 +144,65 @@ def find_latest_awaiting_record_by_ids(self, course_id, program_id) -> dict | No
...
         return latest_record
 
+    IN_FLIGHT_STATUSES = (
+        "PENDING",
+        "IN_PROGRESS",
+        "AWAITING_COMPLETION",
+        "AWAITING_COMPLETION_FAILED",
+    )
+
+    def find_in_flight_records_for_pairs(self, pairs: list[tuple]) -> dict:
```
FabianLCH: Why was this function designed to take a list of `(course_id, program_id)` pairs? It adds unnecessary extra processing of both the input data and the data returned by DynamoDB

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
```diff
@@ -130,13 +144,65 @@ def find_latest_awaiting_record_by_ids(self, course_id, program_id) -> dict | No
...
+
+        "In-flight" means the record exists in any status that hasn't been delivered
+        to Laravel and deleted yet (PENDING, IN_PROGRESS, AWAITING_COMPLETION,
+        AWAITING_COMPLETION_FAILED).
+
+        Single GSI scan per status; in-memory filter for the requested pairs.
+        """
+        result = {pair: None for pair in pairs}
+        if not pairs:
+            return result
```
FabianLCH: This check should go first, and you could simply return `None` rather than generating a set of tuples that map to `None`.

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
```diff
@@ -130,13 +144,65 @@ def find_latest_awaiting_record_by_ids(self, course_id, program_id) -> dict | No
...
+
+        # Normalize to str so callers can pass int/str/Decimal interchangeably.
+        pair_set = {(str(c), str(p)) for c, p in pairs}
+        normalized_to_original = {(str(c), str(p)): (c, p) for c, p in pairs}
```
FabianLCH: We should avoid trying to support as many types as possible and rather have the function expect one type that works and leave converting to that type to the caller. This just makes things more confusing unnecessarily

python/services/lo_mapping_service/app/services/process_batch_transform_results.py
```diff
@@ -3,6 +3,13 @@
 import json
 import os
 import re
+from decimal import Decimal
+from pathlib import Path
+from dotenv import load_dotenv
+
+_SERVICE_ROOT = Path(__file__).resolve().parents[2]
+load_dotenv(_SERVICE_ROOT / ".env", override=False)
```
FabianLCH: No need to manually load values from the environment. The Pydantic settings defined in `app/core/config.py` already does this and should be used across the service as the single source of truth.

python/services/lo_mapping_service/app/services/process_batch_transform_results.py
```diff
@@ -239,15 +257,20 @@ async def process_records(records: list) -> None:
...
                 continue
 
             output_s3_uri = record.get("output_s3_path")
             if not output_s3_uri:
+                # Mirror what create_request now does: <OUTPUT_S3_URI>/<input_filename>.out
                 input_s3_path = record.get("input_s3_path", "")
-                output_s3_uri = input_s3_path.replace("batch_inputs", "batch_outputs") + ".out"
-                if not output_s3_uri:
-                    logger.warning("Record '%s' is missing output_s3_path — using '%s'.", record_id, output_s3_uri)
+                output_prefix = (os.getenv("OUTPUT_S3_URI") or "").rstrip("/")
```
FabianLCH: Processing like this should be handled by the Pydantic setting initialization

python/services/lo_mapping_service/app/services/process_batch_transform_results.py
```diff
@@ -260,12 +283,15 @@ async def process_records(records: list) -> None:
...
             logger.error(
                 "Failed to process record '%s': %s — skipping.",
                 record_id, e,
             )
+            failed.append(record_id)
             continue
 
-    logger.info("Finished processing %d record(s).", len(records))
+    logger.info("Finished processing %d record(s). Succeeded=%d Failed=%d", len(records), len(succeeded), len(failed))
+    return {"succeeded": succeeded, "failed": failed}
```
FabianLCH: Even though we're returning the lists of `succeeded` and `failed` records, it looks like this function's return values are not being used anywhere. Maybe we could just keep the counts for logging?

python/services/lo_mapping_service/deploy/DEPLOY.md
```diff
@@ -0,0 +1,279 @@
+# Note: this documentation was compiled by an AI Tool
+
+# Lambda Deployment Guide
+
+End-to-end PowerShell commands to deploy both Lambda functions for the lo_mapping_service.
+
+## Prerequisites
+
+- AWS CLI installed and configure`d (`aws configure`).
```
FabianLCH: Markdown formatting here is broken

python/services/lo_mapping_service/deploy/DEPLOY.md
FabianLCH, line 1: Rather than having this long markdown file with a lot of segmented powershell code, it would be much better to have a deploy script e.g. written in Python so devs can work off of that and the script is not OS dependent.

python/services/lo_mapping_service/tests/e2e/cleanup.py
FabianLCH, line 1: Having to create files like this one is also a good reason to avoid running E2E tests from the context of `lo_mapping_service`. Lots of hardcoding and SQL to run from a completely separate DB connection initiated by the script. Not good for maintainability

laravel/tests/Feature/SyllabiTest.php
```diff
@@ -181,18 +181,27 @@ public function test_syllabus_duplicate(): void
...
-        $user = User::where('email', 'test-syllabi@ubc.ca')->first();
+        // syllabus.download is POST /syllabi/{syllabusId}/{ext}. user2 is the
+        // current owner (test_syllabus_transfer) so the hasAccess middleware
+        // lets the request through.
+        $user2 = User::where('email', 'test-syllabi-collab@ubc.ca')->first();
         $syllabus = Syllabus::where('course_title', 'Intro to Greatness')->orderBy('id', 'DESC')->first();
-        $response = $this->actingAs($user)->get(route('syllabus.download', $syllabus->id, 'word'))->assertStatus(200);
+
+        $originalCwd = getcwd();
+        chdir(public_path());
```
FabianLCH: Why is this here? There should be no need to change directories in this test

laravel/resources/views/layouts/app.blade.php
```diff
@@ -163,7 +153,7 @@ function gtag(){dataLayer.push(arguments);}
         </div>
     </div>
         <div style="width:100%;">
-            <iframe src="{{ asset('footer.html') }}" width="100%" scrolling="no" style="border:none; margin-bottom:-30px; min-height:426px; max-height: 821px;"/>
+            <iframe src="{{ asset('footer.html') }}" width="100%" scrolling="no" style="border:none; margin-bottom:-30px; min-height:426px; max-height: 821px;"></iframe>
```
FabianLCH: Didn't notice this before, but we should probably load the footer along with the rest of the page HTML instead of via an iframe. Not too sure why it was done this way before!

FabianLCH (changes requested):
Made some good progress, but the changes (mainly for the `lo_mapping_service` need some work. I've covered some of it in my comments, but I wanted to highlight a few things:

- E2E tests are initiated from within the context of `lo_mapping_service`, which:
    - Makes testing changes to the Laravel app that work in conjunction with the full system a lot less intuitive and inflexible
    - Requires a lot of hard-coding and making assumptions about the dev's setup, which worsens dev experience
- The current setup mixes unit tests with E2E tests which make the setup unnecessarily more complicated
    - I know changes were made to ensure they don't conflict when running tests, but it's still not a great setup
- E2E tests should be done from the Laravel app's perspective, using something like Dusk to more closely emulate an actual end-user scenario
    - In the setup, we could then have configuration to automatically start the dependent Python services in e.g. a testing mode that uses LocalStack resources

Also, there's no issue with using AI to help you write code and documentation, however, please make sure to always check the output for quality and readability. Even if the code does what it needs to, not ensuring it's readable makes it hard for other people working on the project to contribute. You should also ensure the solution makes sense and doesn't make unreasonable assumptions about the system it will run on.

laravel/app/Http/Controllers/CourseProgramController.php
Apps247: Initially, the CLO-PLO mapping table for a course isn't shown on the mapping page. The `manual_map_status` field was used in the frontend to decide whether or not to show the mapping table (id `#ManualMapBody`), so it needs to be set to true, in case the user hasn't started manually mapping, but has requested AI suggestions.

We could rename this to `mapping_started` or similar and refactor for clarity

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
Apps247: When the user loads the PLO-CLO mapping page in the Course Wizard, it first checks if there are any AI suggestions in flight for **any of the programs in that course**. 

~For example, If the course is mapped to 3 programs, then this will send an in-flight check request for `[(c0, p0), (c0, p1), (c0, p2)]`, rather than sending 3 separate requests to the backend~

~But after I click AI suggestions for a course-program pair, then it polls for that course-program pair only. If I click multiple, it sends multiple requests with one pair each currently.~

~Should we separate the initial load poll logic from the logic that periodically polls after requesting AI suggestions for a course?~

Edit: 

You're right, we can just accept a single course ID and a list of program IDs. But in the future, if we want to view all the mappings for a program instead of for a course, it's worth keeping the current set up. If not, happy to change it to just accept 1 course ID and a list of program IDs. 

And since the 'batch' poll is only happening on the initial load of the page, we can even separate the initial load poll logic from the logic that periodically polls after requesting AI suggestions for a course

laravel/tests/Feature/SyllabiTest.php
Apps247: The code that the test calls needs the syllabus template docx files in `laravel/public`. If we remove these lines, the test fails because laravel tests don't start the web server by default (which would serve `/public`)

We could make all the tests run from `laravel/public` by default?
Or we could test the syllabus download in the E2E tests and remove it from here?

laravel/resources/views/layouts/app.blade.php
Apps247: Seems like it was done because the footer uses the footer stylesheet provided by UBC, and loading it in the page directly makes it conflict with the styles we already have.
But I tried adding the stylesheet in and loading the footer directly, and it's only minor spacing and font differences

laravel/tests/Feature/SyllabiTest.php
FabianLCH: Hm, I had a look at the underlying code (`SyllabusController > download(...)`). I don't think it should be grabbing the template from the `/public` directory.

Can you try moving the templates to `laravel/resources` and using `resource_path(...)` to get the path to the corresponding templates? That way, the tests and app shouldn't depend on the `laravel/public` directory being served.

laravel/resources/views/layouts/app.blade.php
FabianLCH: Noted! There should be no problem with the minor differences

laravel/app/Http/Controllers/CourseProgramController.php
FabianLCH: I see. Maybe we can modify the frontend code to check for one of `manual_map_status` or `ai_suggestion_status` to show the mapping table, since now either of those two should be the condition for us to show it.

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
FabianLCH: Good point, let's keep the current set up as I feel there's a good chance we'll want to do the reverse operation in the future. Also, I think separating the batch poll from the periodic poll logic is a good idea!

python/services/lo_mapping_service/tests/e2e/cleanup.py
Apps247: The test fastapi service now handles DynamoDB side cleanup, and composer handles laravel side cleanup

laravel/app/Http/Controllers/CourseProgramController.php
Apps247: Same reasoning as [this comment](https://github.com/forestryTLS/curriculum-development-tool/pull/16#discussion_r3299346136)

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
Apps247: On second look, the DynamoDB table doesn't have an index on CLO id and PLO id. Unless we update the index, I don't think it's worth separating them, but happy to do so if you think it's beneficial on next review!

Apps247:
**Summary of changes since last PR**
- Bootstrap v5 syntax for `handleUpdate()` function
- AI Suggestions page frontend minor tweaks
composer test spawns separate db and reseeds when tests are run
the deploy script for the AI Suggestions lambdas and EventBridge trigger to AWS
- Laravel 10 > 11 upgrade, removed backpack
- FastAPI test server that also spawns a LocalStack container and has test routes
- Pest browser tests (in `laravel/tests/Browser/AISuggestionsE2E`. Also has a `README.md` in the same folder. 
   - The bottom of this `README.md` has a Notes section with some design quirks. Please let me know if you think I should change my approach on any of these
- Non-AI-suggestions-feature bug fixes
   - From your feedback: syllabus template location, footer iframe, handling program mapping scale changing after mapping
   - `validateMaxLength` was not working in step1

laravel/resources/views/courses/courseCollabs.blade.php
```diff
@@ -284,7 +284,7 @@ class="btn btn-secondary btn-sm"
                 $(this).addClass('was-validated');
             }
             // readjust modal's position/height
-            document.querySelector('#addCourseCollaboratorsModal' + courseId).handleUpdate();
+            bootstrap.Modal.getInstance(document.querySelector('#addCourseCollaboratorsModal' + courseId))?.handleUpdate();
```
FabianLCH: Hm, I'm thinking we can standardize this (and other instances) by using the jQuery form `$("...").modal("handleUpdate")` (ran into the same issue while reviewing Kabir's additions and suggested that form since it's simpler)

laravel/resources/views/programs/wizard/header.blade.php
```diff
@@ -533,18 +533,16 @@ function campusOtherOption() {
...
+    function validateMaxlength(e){
+        //Whitespaces are counted as 1 but character wise are 2 (\n).
+        var MAX_LENGTH = event.target.getAttribute("maxlength");
+        var currentLength = event.target.value.length;
+        var whiteSpace = event.target.value.split(/\n/).length;
+        if((currentLength+(whiteSpace))>MAX_LENGTH)
+        {
+            //Goes to MAX_LENGTH-(whiteSpace)+1 because it starts at 1
+            event.target.value = event.target.value.substr(0,MAX_LENGTH-(whiteSpace)+1);
         }
```
FabianLCH: Let's update this to use the event variable `e` explicitly passed to the function instead of the global `window.event` (which is fragile and not very reliable!)

laravel/tests/Browser/AiSuggestionsE2E/MainWorkflowTest.php
```diff
@@ -0,0 +1,180 @@
...
+        ->click('[data-bs-target="#createCourseModal"]')
+        ->type('#course_code', 'HAPP')
+        ->type('#course_num', '101')
+        ->type('#course_title', 'E2E Normal workflow Course')
+        ->select('#course_semester', 'W1')
+        ->select('#course_year', '2026')
+        ->select('#delivery_modality', 'O')
+        ->select('#standard_category_id', '1')
+        ->wait(1)
+        ->pressAndWaitFor('#createCourse #submit', 2); // POSTs and redirects to /courseWizard/{id}/step8
```
FabianLCH: Maybe we can add an `assertUrlIs(...)` so it's easier to notice major changes to the course wizard flow in the future after running tests

laravel/tests/Browser/AiSuggestionsE2E/MainWorkflowTest.php
```diff
@@ -0,0 +1,180 @@
...
+        ->pressAndWaitFor('#saveCLOChanges button:has-text("Save Changes")', 1);
+
+    $clo = LearningOutcome::where('course_id', $course->course_id)->firstOrFail();
+
+    // Create program
+    $page->click('a[href$="/home"]')
+        ->pressAndWaitFor('[data-bs-target="#createProgramModal"]', 0.5)
+        ->type('#program', 'E2E Normal workflow Program')
+        ->check('input[name="level"][value="Bachelors"]')
+        ->pressAndWaitFor('Add', 1); // redirects to /programWizard/{id}/step1
```
FabianLCH, line 48: Let's add an `assertUrlIs(...)` here too for the redirect!

laravel/tests/Browser/AiSuggestionsE2E/ScaleChangeCleanupTest.php
```diff
@@ -0,0 +1,168 @@
...
+    $this->actingAs($user);
+
+    // Mock SageMaker request submission and completion
+    putPendingRecord($course->course_id, $programAi->program_id);
+    setAwaitingCompletion($course->course_id, $programAi->program_id, [
+        ['clo_id' => $clo->l_outcome_id, 'plo_id' => $ploAi1->pl_outcome_id, 'labels' => ['I']], // icon on I
+        ['clo_id' => $clo->l_outcome_id, 'plo_id' => $ploAi2->pl_outcome_id, 'labels' => ['A']], // icon on A
+    ]);
+
+    $page = visit("/courseWizard/{$course->course_id}/step5");
```
FabianLCH: Should this use the `visit_v` with the set viewport size you defined?

laravel/tests/Browser/AiSuggestionsE2E/MultipleProgramsIndependentTest.php
```diff
@@ -0,0 +1,103 @@
...
+    putPendingRecord($course->course_id, $program1->program_id);
+    putPendingRecord($course->course_id, $program2->program_id);
+
+    $this->actingAs($user);
+
+    $options1 = "#mappingOptions-{$course->course_id}-{$program1->program_id}";
+    $options2 = "#mappingOptions-{$course->course_id}-{$program2->program_id}";
+
+    // Check that both programs show the waiting state
+    $page = visit("/courseWizard/{$course->course_id}/step5");
```
FabianLCH: Should this be `visit_v` here?

laravel/package.json
```diff
@@ -20,7 +20,8 @@
         "vue": "^3.2.25"
     },
     "dependencies": {
-        "@popperjs/core": "^2.11.4"
+        "@popperjs/core": "^2.11.4",
+        "playwright": "^1.60.0"
```
FabianLCH: Playwright should be a dev dependency!

python/services/lo_mapping_service/app/test.py
```diff
@@ -0,0 +1,123 @@
...
+            CreateBucketConfiguration={"LocationConstraint": region},
+        )
+        print(f"[test] Created S3 bucket {bucket}")
+    except s3.exceptions.BucketAlreadyOwnedByYou:
+        print(f"[test] S3 bucket {bucket} already exists, reusing")
+
+
+def reset_dynamodb():
+    """Drop the LO mapping requests table if it exists."""
+    load_dotenv()
```
FabianLCH: Won't this make the script override the explicitly set test env values by loading the actual values stored in the `lo_mapping_service/.env` file? I'm guessing what you'd want to do is define a shared `settings` object with the test env variables set at definition (e.g. `Settings(ACCESS_KEY="...", ...)`) and then pass that `settings` object to `LOMappingRequestDynamoDBRecord` below so it uses the LocalStack DynamoDB table

python/services/lo_mapping_service/deploy/deploy_script.py
FabianLCH, line 1: I'm thinking we could add a tag like `{ "AppName": "curriculum-development-tool" }` to the resources created here to make it easier to track them on AWS (especially when looking at cost, resource usage, etc.)

FabianLCH, line 1: I know the `LOMappingRequestDynamoDBRecord` store already handles table creation, but I think it would be good to add a step to this script that creates the DynamoDB table. That way, we can have all the resources ready, plus it would serve as additional documentation for the table's definition (speaking of which, we should probably document that somewhere if not already done)

python/services/lo_mapping_service/deploy/deploy_script.py
```diff
@@ -0,0 +1,231 @@
...
+        "HF_IMAGE_URI",
+        "OUTPUT_S3_URI",
+        "IAM_ROLE_ARN",
+        "LAMBDA_ROLE_ARN",
+    ):
+        print(f"  {key} = {getattr(settings, key)}")
+    return settings
+
+
+def zip_handler(module_name: str, dest: Path) -> Path:
```
FabianLCH, line 86: Let's add a note here that, since our lambda handlers only use standard library functions and boto3, which is automatically included in the AWS Lambda python runtime, packaging only requires the `.py` scripts themselves. However, in the future, if external libraries are needed, this part of the deploy script will need to be modified, and additional steps will be required to package those external libraries as well

FabianLCH (changes requested):
The changes are looking pretty good! I really like the new setup for the E2E tests with Pest. I mostly flagged some fairly minor changes I think we should make.

Below are some general comments/suggestions:
- I'm thinking we could add a test for the case where a user deletes a CLO or PLO while the AI suggestion request is in flight. In such case, when the request completes and Laravel gets the results, it should still store the AI suggestions for the remaining CLO-PLO combinations that were deemed as 'mapped' by the LLM
- Currently, it seems the app does not automatically save the AI suggestions as mappings on the database, so the user still has to click through the checkboxes in the UI to use the suggested mappings. Let's change this so the app automatically does this!
    - We'll also have to add/modify tests to reflect this change

Apps247:
**Summary of Changes Made since last PR**
- When AI Suggestions are received, made unmapped manual rows set to the AI Suggestions by default 
    - Added Unsaved changes warning on Program Outcome Mapping page, and prevented autoreload to load AI Suggestions if there are unsaved changes.
    - Added dedicated test for the manual maps defaulting to AI Suggestions, `ManualMapAIDefaultsTest.php`, and added assertions in `MainWorkflowTest.php`. 
    - I didn't add manual map assertions to the other tests, as I felt it would take away from their focus and slow them down, but I'm happy to add them if it's required.
- Made `storeAISuggestions()` skip over rows where the CLO or PLO were deleted _after_ sending the request. This prevents the whole commit from being aborted due to Foreign Key constraints. 
    - Added test for deleting CLO or PLO when request is in flight, `DeleteCLOPLOInFlightTest.php`.
- Made the LO Mapping Service `deploy_script.py` create S3 bucket, DynamoDB table, and tag all resources created with an `AppName` that can be set in the service's `.env`. Similarly made the lambda that creates the SageMaker job,  `lambda_handler_start_batch_tranform_job.py` , tag the job.
- Other minor fixes pointed out in the comment thread

python/services/lo_mapping_service/app/services/lo_mapping_request_dynamo_db_request.py
FabianLCH: All good, let's keep as is then!

laravel/resources/views/courses/wizard/step1.blade.php
```diff
@@ -246,24 +246,24 @@ class="form-control @error('clo_shortphrase') is-invalid @enderror"  form="saveC
...
+                // reset form
                 $(this).trigger('reset');
                 $(this).removeClass('was-validated');
             } else {
                 // mark form as validated
                 $(this).addClass('was-validated');
             }
             // readjust modal's position
-            bootstrap.Modal.getInstance(document.querySelector('#addLearningOutcomeModal'))?.handleUpdate();
+            $('#addLearningOutcomeModal').modal("handleUpdate")();
```
FabianLCH: You can remove the function call syntax at the end here as it is invalid! `$('...').modal("handleUpdate")` calls the method and does not return a callable object, so having `();` at the end results in an error

laravel/resources/views/courses/wizard/step2.blade.php
```diff
@@ -281,7 +281,7 @@
                 $(this).addClass('was-validated');
             }
             // readjust modal's position
-            bootstrap.Modal.getInstance(document.querySelector('#addAssessmentMethodModal'))?.handleUpdate();
+            $('#addAssessmentMethodModal').modal("handleUpdate")();
```
FabianLCH: Same as the comment for `step1.blade.php` (also applies for any other instances in other blade files)

laravel/tests/Browser/AiSuggestionsE2E/DeleteCLOPLOInFlightTest.php
```diff
@@ -0,0 +1,86 @@
...
+    // Should also have set the manual map to AI Suggestions
+    $this->assertDatabaseHas('outcome_maps', [
+        'l_outcome_id' => $clo2->l_outcome_id,
+        'pl_outcome_id' => $plo2->pl_outcome_id,
+        'map_scale_id' => $I,
+    ]);
+
+    // Deleted outcomes should not be re-added to database
+    $this->assertDatabaseMissing('outcome_map_ai_suggestions', ['l_outcome_id' => $clo1->l_outcome_id]);
+    $this->assertDatabaseMissing('outcome_map_ai_suggestions', ['pl_outcome_id' => $plo1->pl_outcome_id]);
```
FabianLCH: Aren't these the exact same checks as in lines 65 & 66? I'm guessing you maybe wanted to check that 'outcome_map' (the manual mappings table) doesn't have the deleted CLO & PLO

FabianLCH (approved):
Just pointed out some very small things that don't affect the app's functionality much, so I'll go ahead and approve this!

I also noticed the package dependency files listed some changes after I ran installs to test things locally. More specifically, I noticed playwright was not listed as a dev dependency in `package-lock.json`, probably means another install wasn't run after it was moved in `package.json`. Also, testcontainers was missing from the LO mapping service's `requirements.txt`. Not super important really!

Apps247:
- Added `testcontainers` to LO Mapping Service `requirements.txt`.
- Updated `package-lock.json`
- In `DeleteCLOPLOInFlightTest.php` made lines 76 and 77 check the `outcome_maps` table instead of `outcome_map_ai_suggestions`
- Removed `()` at the end of modal `handleUpdate` calls.

- Added `hover()` calls in 2 test sections with flaky button presses.

FabianLCH (commented):
Thanks for making those last changes, Aprameya! Getting this merged

PR #17, opened 2026-06-05
Course topics and materials support + syllabi extraction

### LINKED ISSUE #13: Course content search across programs
# Description

Currently, the Curriculum Development Tool does not implement any search functionality. To view the content covered by a program, a user must access each course in the program individually and review its description, assessment methods, etc. Additionally, the information the tool stores is limited.

To facilitate understanding of topic coverage across courses and programs, the tool should allow users to perform system-wide searches of programs based on one or more properties related to course content.

# High-Level Tasks

1. Expand syllabus parsing functionality to extract and store additional course information (e.g. course topics)
2. Design and build functionality to allow performing system-wide searches of courses/programs
3. Expand access management functionality to restrict searchable content according to user role and memberships
 
# Guiding Questions

### What kind of additional information should the tool extract from syllabi (if available)?
- Main course topics (e.g. from the course schedule, lectures, modules, etc.)
- Course materials (e.g. videos, papers, etc.)
    - Name
    - Type
    - Description

### What is the main use case for this feature?
During curriculum development meetings, instructors gather to manually go over their learning materials one by one and attempt to identify gaps/overlaps in course content. Organizing these meetings requires considerable resources and time investment.

Additionally, these meetings tend to focus on _core_ courses. Thus, _electives_ are not always thoroughly evaluated.

Two key goals of this feature are:
1. Compile and present courses/programs that match search queries related to course properties
2. Help instructors and program designers answer **basic** questions about course/program content coverage

For example, a user could search for "climate change" and get a list of courses that mention the term in their description, learning objectives, topics, etc.

### Are there properties that should be prioritized over others?
If a search property is not specified, matches should be prioritized according to the properties listed below:
1. Topics
2. Learning Objective
4. Assessment Activities
5. Course Descriptions

### What search statistics would be most beneficial?

- Number of matching courses per property
    - This could be inspired by the "X results" usually included in search engine result pages. For example:
        - "Found in 3 course descriptions, 7 learning objectives, 1 assessment activity"
- Number of courses matching the search
- Specific courses matched (perhaps a list of course codes and numbers)

Additionally, the feature could offer the option to filter results to a specific course instead of a system-wide search when necessary.

NEOGDAWG:
- Added course topics and course materials end-to-end including migrations, schema constraints, models, factories, controllers, routes, wizard UI steps, navigation updates, and summary page display.
- Implemented create/delete functionality for course topics and course materials in the course wizard, and added feature tests to cover those workflows.
- Expanded the syllabi parser to return topics and materials, and implemented extraction logic for topic sections, topic tables, text fallback topics, material sections, bullet/list materials, citations, cleanup, deduplication, and material type inference.
- Expanded test suite for syllabi parsing; created and added test syllabi covering different topic/material formats, including tables, non-table lists, alternate headings, messy material sections, multiple material types, missing topic keywords and multi page topic tables.
- wired up material and topics parsing to Laravel app and Database so now topics and material parsing works end-to-end in the app
- manually tested 30+ real course syllabi and reiterated to improve implementation of parser to fit more edge cases and variety of formats
- Updated course creation from uploaded syllabi to persist extracted topics and materials into the new database tables.
- Restricted automatic course creation from syllabi uploads to PDF files only, since topic/material extraction currently relies on PDF parsing.
- Topic/material parsing is heuristic-based, so outputs may include some broad search tags or minor noise, but the goal is useful searchable metadata rather than perfect syllabus reconstruction.
- Add support for syllabus-specific patterns: including alternate section headings, wrapped PDF table cells, continued multi page tables, module/theme/topic-column detection, citation-style materials, generic Canvas/resource filtering, and broader topic tags for search.


laravel/app/Http/Controllers/CourseMaterialController.php
```diff
@@ -0,0 +1,185 @@
...
+                    $newCourseMaterial->name = $newmat['name'];
+                    $newCourseMaterial->type = $newmat['type'] ?? null;
+                    $newCourseMaterial->description = $newmat['description'] ?? null;
+                    $newCourseMaterial->is_required = isset($newmat['is_required']);
+                    $newCourseMaterial->url = $newmat['url'] ?? null;
+                    $newCourseMaterial->course_id = $courseId;
+                    $newCourseMaterial->position = $index + 1;
+                    $newCourseMaterial->save();
+                }
+            }
```
FabianLCH, line 118: A minor optimization we could do in case there are lots of materials to save to the DB is to build an array with the values to insert and use `CourseMaterial::insert($values)` to do bulk creation

laravel/app/Http/Controllers/CourseMaterialController.php
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * @param  \App\Models\CourseTopic  $courseTopic
+     */
+    public function destroy(Request $request, $course_material_id): RedirectResponse
+    {
+        //
+        $courseMaterial = CourseMaterial::where('course_material_id', $course_material_id)->first();
+        $course_id = $request->input('course_id');
+
+        if ($courseMaterial && $courseMaterial->delete()) {
```
FabianLCH, line 184: Let's add an extra check here to ensure the provided `course_id` matches `$courseMaterial->course_id` (i.e. the material belongs to the selected course)

laravel/app/Http/Controllers/CourseTopicController.php
```diff
@@ -0,0 +1,166 @@
...
+                        continue;
+                    }
+
+                    $newCourseTopic = new CourseTopic;
+                    $newCourseTopic->topic = $newTopic;
+                    $newCourseTopic->course_id = $courseId;
+                    $newCourseTopic->position = $index + 1;
+                    $newCourseTopic->save();
+                }
+            }
```
FabianLCH, line 100: Can also do a bulk insert here!

laravel/app/Http/Controllers/CourseTopicController.php
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @param  \App\Models\CourseTopic  $courseTopic
+     */
+    public function destroy(Request $request, $course_topic_id): RedirectResponse
+    {
+        //
+        $courseTopic = CourseTopic::where('course_topic_id', $course_topic_id)->first();
+        $course_id = $request->input('course_id');
+
+        if ($courseTopic && $courseTopic->delete()) {
```
FabianLCH, line 167: Similar to course materials, we can add a check here to ensure the topic belongs to the specified course

laravel/app/Jobs/ProcessCourseSyllabiFile.php
```diff
@@ -115,6 +117,30 @@ public function handle(): void
...
+        foreach($courseMaterials as $cm){
+            if(!empty($cm)){
+                $courseMaterial = new CourseMaterial();
+                $courseMaterial->course_id = $course->course_id;
+                $courseMaterial->name = $cm->name;
+                $courseMaterial->type = $cm->type;
+                $courseMaterial->description = $cm->description;
+                $courseMaterial->is_required = false; #default would be false but user can change it if needed (more material found in syllabi are optional than required)
+                $courseMaterial->save();
+            }
```
FabianLCH, line 173: We can also try a bulk insert for course topics and materials here!

laravel/database/migrations/2026_06_02_000001_change_course_material_name_to_text.php
FabianLCH, line 1: Is this migration necessary? It looks like you already defined the `name` column as a text column

laravel/database/migrations/2026_06_02_000002_change_course_topic_topic_to_text.php
FabianLCH, line 1: Same comment as the previous migration

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -0,0 +1,308 @@
...
+                addCourseMaterial();
+                // reset form
+                $(this).trigger('reset');
+                $(this).removeClass('was-validated');
+            } else {
+                // mark form as validated
+                $(this).addClass('was-validated');
+            }
+            // readjust modal's position
+            document.querySelector('#addCourseMaterialsModal').handleUpdate();
```
FabianLCH: It looks like this won't work because you are querying the modal's underlying `<div>` HTML element (which does not have a `handleUpdate` method) and not the Bootstrap modal object itself. I think what you want (using jQuery) is the following:
```suggestion
            $('#addCourseMaterialsModal').modal("handleUpdate");
```

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -0,0 +1,308 @@
...
+                                <h5 class="modal-title" id="addCourseMaterialsModalLabel"><i class="bi bi-pencil-fill btn-icon mr-2"></i> Course Materials</h5>
+                            </div>
+
+                            <div class="modal-body">
+                                <form id="addCourseMaterialsForm" class="needs-validation" novalidate>
+                                    <!-- Materials have multiple fields, so the add form collects one complete material row. -->
+                                    <div class="row justify-content-between align-items-end m-2">
+                                        <div class="col-4">
+                                            <label for="courseMaterialName" class="form-label fs-6"><b>Name</b></label>
+                                            <input id="courseMaterialName" class="form-control" oninput="validateMaxlength()" onpaste="validateMaxlength()" maxlength="191" placeholder="Material name" required>
```
FabianLCH: Looks like `validateMaxlength` is not defined (I kept seeing errors while doing some manual testing)!

laravel/resources/views/courses/wizard/step9.blade.php
```diff
@@ -0,0 +1,220 @@
...
+                addCourseTopic();
+                // reset form 
+                $(this).trigger('reset');
+                $(this).removeClass('was-validated');
+            } else {
+                // mark form as validated
+                $(this).addClass('was-validated');
+            }
+            // readjust modal's position 
+            document.querySelector('#addCourseTopicsModal').handleUpdate();
```
FabianLCH: Same as comment on `step10.blade.php`, you can use:
```suggestion
            $('#addCourseTopicsModal').modal("handleUpdate");
```

laravel/tests/Feature/CourseTest.php
FabianLCH, line 1: Let's add some assertions to the tests here to ensure users are correctly redirected after successful POST requests

FabianLCH, line 1: Let's also add some tests to ensure the server responds accordingly when required fields are not submitted! One good reason for this is that users can modify client-side form verification. Another is, someone else may modify the schema in the future, and such tests serve as an indicator of an important functionality change that should be properly addressed

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+    if(table_topics):
+        return clean_topics(table_topics) #clean and filter all the topics 
+    
+
+    text_topics = get_topics_from_text(doc_without_header, topic_pages) 
+    return clean_topics(text_topics)
+
+
+
+def find_topic_section(doc: list[str]) -> list[int]:
```
FabianLCH: Let's add a docstring to this function explaining it returns the indices of the pages where a topic section is found, and that a topic section can span across multiple pages. Also, maybe we can rename the function to something like `find_topic_section_page_indices` to make its purpose clearer at first glance

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+                topic_pages.append(page_index)
+                in_section = True
+                break
+
+        if in_section:
+            break
+
+
+    if in_section:
+        for next_page_index in range(topic_pages[0] + 1, len(doc)):
```
FabianLCH, line 1294: Might be wrong, but if the section stop heading was in the same page as the start heading page (topics are all contained within the same page), this code would skip it, resulting in more page indices returned than necessary

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+                if clean_line in TOPIC_SECTION_STOP_HEADINGS:
+                    return topic_pages
+
+    return topic_pages
+
+
+#if a line looks like a start of a new syllabus section, return true
+def is_section_boundary(line: str, stop_headings: set[str], topic_headings: set[str]) -> bool:
+    stripped_line = line.strip()
+    clean_line = normalize_section_line(stripped_line)
```
FabianLCH, line 1319: I think we can get rid of `stripped_line` here since `normalize_section_line` already has a step to strip whitespace

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+    if not clean_line:
+        return False
+    if clean_line in topic_headings:
+        return False
+    if clean_line in stop_headings:
+        return True
+
+    clean_without_number = normalize_section_line(clean_line)
+    if clean_without_number in stop_headings:
+        return True
```
FabianLCH: Unless I'm missing something, wouldn't `clean_without_number` be the same as `clean_line` and result in repeating the `if clean_line in stop_headings: ...` check?

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+    words = clean_without_number.split()
+    if len(words) > 6:
+        return False
+
+    if stripped_line.isupper() and len(words) >= 1:
+        return True
+
+    if re.match(r"^\d+(\.\d+)*[.)]?\s+[A-Z][A-Za-z&/ -]+$", stripped_line):
+        return True
```
FabianLCH: Let's add some comments explaining here why each check defines whether the line is or isn't a section boundary

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+        tables = doc[page_index].find_tables()
+        if not tables:
+            continue
+
+        for table in tables.tables:
+            extracted_table = table.extract()
+            if not extracted_table or len(extracted_table) < 2:
+                if extracted_table and len(extracted_table) == 1:
+                    previous_header = [normalize_table_heading(cell or "").lower() for cell in extracted_table[0]] # Some PDFs split the header row from the table body across pages
```
FabianLCH: I don't think you need the extra `.lower()` call since `normalize_table_heading` already does it

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+            if not (set(header) & (topic_column_headings | title_column_headings | module_topic_column_headings | theme_topic_column_headings)) and previous_header and len(previous_header) == len(header):
+                header = previous_header # use a remembered prev header when PymuPDF extracts the body rows as a separate table
+
+
+            header_set = set(header)
+            has_topic_header = bool(header_set & topic_column_headings)
+            has_assessment_header = bool(header_set & assessment_table_headings)
+            table_context_headings = set()
+            for row in cleaned_table[:12]:
```
FabianLCH, line 1390: Why is the looping limited to the first 12 rows specifically?

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+    if stripped_line.isupper() and len(words) >= 1:
+        return True
+
+    if re.match(r"^\d+(\.\d+)*[.)]?\s+[A-Z][A-Za-z&/ -]+$", stripped_line):
+        return True
+
+    return False
+
+
+def get_topics_from_table(doc: pymupdf.Document, topic_pages: list[int]) -> list[str]:
```
FabianLCH, line 1341: I feel like this really long function should be split into a couple of smaller sub-functions as there are a lot of nested loops and conditionals that make it hard to follow

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+    for row in rows[:6]:
+        if column_index >= len(row):
+            continue
+        cell = re.sub(r"\s+", " ", row[column_index]).strip()
+        if not cell:
+            continue
+
+        non_empty_count += 1
+        lower_cell = cell.lower()
+        word_count = len(re.findall(r"[A-Za-z]{3,}", cell))
```
FabianLCH: This could be made a bit more robust by checking for matches of 2 or more characters, including accented characters and adding a pattern for hyphenated words (which I believe should count as one word)

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+        looks_bad = (
+            re.fullmatch(r"(day|week|module|unit)\s*[\divxlcdm/ -]+", lower_cell, re.IGNORECASE) or
+            re.search(r"\b\d{1,2}:\d{2}\b", lower_cell) or
+            re.search(r"\b(mon|tue|wed|thu|fri|monday|tuesday|wednesday|thursday|friday|jan|feb|mar|apr|may|jun|jul|aug|sep|sept|oct|nov|dec)\b", lower_cell) or
+            re.search(r"\b(canvas|reading|readings|assignment|assignments|due|quiz|exam|test|prep|slides?|http|www|room|location|mark|weight)\b", lower_cell) or
+            re.search(r"\b(?:\(\d{4}\)|\d{4})\.", cell)
+        )
```
FabianLCH, line 1671: It'd be good to add comments here briefly explaining what "bad" data each pattern represents

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+#to decide if a given cell should become one topic or multiple topics
+def split_topic_cell(cell: str) -> list[str]:
+    lines = []
+
+    for item in cell.split("\n"):
+        item = item.strip()
+        if not item:
+            continue
+
+        clean_item = re.sub(r"\s+", " ", item.lower()).strip()
```
FabianLCH: Maybe we can use `re.compile` for this since I've noticed a couple of places use the same pattern to remove consecutive spaces

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+            continue
+
+        lines.append(item)
+
+    if not lines:
+        return []
+
+    numbered_or_bulleted_lines = []
+    for item in lines:
+        if re.match(r"^(\d+[\.)]|[-•*])\s+", item):
```
FabianLCH: Probably should `re.compile` this one as well since it's reused below (and can provide some minor performance gains when the same regular expression is used often)

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+            numbered_or_bulleted_lines.append(item)
+
+    if len(numbered_or_bulleted_lines) > 1:
+        split_topics = []
+        current_topic = ""
+
+        # only split when the cell clearly contains a list - plain newlines in PDF tables are usually just wrapped text
+        for item in lines:
+            starts_new_topic = re.match(r"^(\d+[\.)]|[-•*])\s+", item)
```
FabianLCH: Since in the code above this block we're already traversing the `lines` list, doing the same regex match and populating the `numbered_or_bulleted_lines` list for the conditional here, why not just skip that and traverse `lines` just once?

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+        current_topic = ""
+
+        #module tables sometimes put two module topics in one cell, while other newlines are just wrapped text
+        for item in lines:
+            if re.match(r"^module\s+\d+:", item, re.IGNORECASE):
+                if current_topic:
+                    split_topics.append(current_topic.strip())
+                current_topic = item
+            else:
+                current_topic = f"{current_topic} {item}".strip()
```
FabianLCH: Like other example, we're traversing all of `lines` in the above block anyway, so I don't think we need to populate `module_lines` just to check its length and loop over `lines` again

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+            continue
+        if re.fullmatch(r"[A-Za-z]?\d+(\.\d+)*[.)]?", topic):
+            continue
+        if re.fullmatch(r"\d+(\.\d+)*[.)]?", topic): #remove if toppic is a number like value
+            continue
+        if re.match(r"^quiz unit \d+$", topic) or re.match(r"^assignment \d+", topic) or re.match(r".*\bdue\b.*", topic):
+            continue
+
+        lower_topic = topic.lower()
+        lower_topic = lower_topic.replace("–", "-").replace("—", "-") #there are diff dashes that python treats differently
```
FabianLCH: Already did this replacement for `topic` above

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+        if not topic:
+            continue
+        if re.fullmatch(r"week\s+\d+\s*-\s*\d+", topic, re.IGNORECASE): 
+            continue
+        if re.fullmatch(r"[A-Za-z]?\d+(\.\d+)*[.)]?", topic):
+            continue
+        if re.fullmatch(r"\d+(\.\d+)*[.)]?", topic): #remove if toppic is a number like value
+            continue
+        if re.match(r"^quiz unit \d+$", topic) or re.match(r"^assignment \d+", topic) or re.match(r".*\bdue\b.*", topic):
+            continue
```
FabianLCH: Similar to the third regex here, please add a short comment to explain what the other patterns are targetting

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+                    paragraph = f"{paragraph} {continuation}".strip()
+                    skip_line_indexes.add(next_index) 
+                    next_index += 1
+
+                recommended_books_match = re.search(r"recommended reading:\s*(.+?\(\d{4}\))\s+and\s+(.+?\(\d{4}\))", paragraph, re.IGNORECASE)
+                if recommended_books_match:
+                    for book_name in recommended_books_match.groups():
+                        append_material(materials, book_name.strip(" ,."), "recommended reading")
+                continue
```
FabianLCH, line 2308: I think we can add/improve comments here to explain the specific parsing for lines that contain `"no required textbook"`

FabianLCH: Can also put this in a separate function to make the code easier to follow

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+                        break
+                    if citation_pattern.match(continuation) or bibliography_citation_pattern.match(continuation):
+                        break
+
+                    citation_name = f"{citation_name} {continuation}".strip()
+                    skip_line_indexes.add(next_index) # bibliography entries often wrap across PDF lines so this keeps one citation as one material.
+                    next_index += 1
+
+                append_material(materials, citation_name, current_description or "Course materials", infer_description=current_description)
+                continue
```
FabianLCH: Similar to the block above, we can add this to a separate function to make the code in this function a bit easier to follow

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+    if "documentary" in name or "documentary" in description:
+        return "video"
+    if name.startswith("vice:"):
+        return "video"
+
+    for t in material_types:
+        if t in name:
+            return t
+    if "websites and online articles" in description and "article" in name:
+        return "article"
```
FabianLCH: Maybe we can move this above with the other specific checks before falling back to checking for instances of the `material_types` in `name`

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+    for material in materials:
+        name = material.get("name", "") 
+        material_type = material.get("type", "")
+        description = material.get("description", "")
+
+
+        name = re.sub(r"\s+", " ", name).strip()
+        material_type = re.sub(r"\s+", " ", material_type).strip()
+        description = re.sub(r"\s+", " ", description).strip()
```
FabianLCH: You can `re.compile` this pattern since it's being reused often

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1066,3 +1163,1230 @@ def encode_assessment_and_weight(assessment_and_weights: list[tuple[str, str]])
...
+
+        name = name.replace("–", "-").replace("—", "-")
+        description = description.replace("–", "-").replace("—", "-")
```
FabianLCH: Since I've seen this replacement across multiple parts of the file, and it's done pretty often, maybe we can convert it to a regular expression replace that targets the two special dash characters and `re.compile` it for a minor performance improvement

python/services/syllabi_service/app/services/syllabus_parser.py
FabianLCH, line 1: A couple of general comments here:

- To make the parsing logic easier to follow, I'd like to see more comments, especially in places where the parsing addresses specific formats you observed that often include specific words, word counts, starting/ending terms, etc.
- To improve in-code docs a bit, let's add docstrings to each function briefly explaining what they do. They can even be one-liners
- There are a few really long functions with lots of nested conditionals and loops. I think we can split them a bit better into sub-functions to improve readability and make it easier for other devs unfamiliar with the logic to make changes going forward
- There's a couple of regular expressions that get reused a lot which can be optimized a bit via `re.compile`

FabianLCH (changes requested):
Changes are looking good so far! This needs a bit more work before we can get it merged, but it should be mostly minor improvements and code cleanup

laravel/app/Http/Controllers/CourseTopicController.php
NEOGDAWG: Done! topics and materials to use bulk inserts in both the controllers and syllabus-processing job.

laravel/app/Http/Controllers/CourseTopicController.php
NEOGDAWG: Added course ID checks to ensure the topics and materials belong to right course

laravel/database/migrations/2026_06_02_000001_change_course_material_name_to_text.php
NEOGDAWG: You are right. I believe I was using string as the type originally, which resulted in me adding redundant migrations to turn it to text. Removed both migrations now!

laravel/resources/views/courses/wizard/step10.blade.php
NEOGDAWG: applied the suggested change manually!

laravel/resources/views/courses/wizard/step10.blade.php
NEOGDAWG: It seemed to be out of the scope. I have moved the method so it should work now. This will probably also fix a bunch of other calls to validateMaxLength

laravel/tests/Feature/CourseTest.php
NEOGDAWG: Done!

NEOGDAWG: I added server side validation to the topic and material controllers, along with tests that confirm missing required fields produce validation errors and do not change DB.

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Done!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: You are right, I added functionality and a test to now check the rest of the page after finding the section heading so it will stop if a stop heading is the in the same page.

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: I have a later fallback check in the code that uses this stripped_line var to check if a line is all caps, as most singular lines that are all caps are section boundaries. I think it would be okay to keep stripped_line in this case since normalize_section_line() removes any capitals.

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: I have removed the duplicate code!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: PyMuPDF sometimes places itinerary headings below the header row, so I sample the first 12 rows without scanning the entire table. 12 is not really a significant number. Let me know if there is anything you recommend doing for this code block!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Good idea, done!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Yeah looks like I overlooked that. I made this section a one pass loop now!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Also made this a one pass!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Good catch. The earlier material_types loop already returns article when it appears in the name so this check is unreachable. I’ll remove it

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Done! I used re.compile for whitespace patterns and the dash patterns since those were used in a lot of places.

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Done!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: I have refactored this function and created helpers and sub functions to make it easier to follow!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: added!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Made into a separate function. Also added small comments.

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: done!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Addressed these now. I added docstrings/comments across the parser, moved repeated regexes and heading sets into shared constants, and split the larger topic/material parsing sections into smaller helpers while keeping the existing behavior unchanged. The Python parser tests are still passing. I removed dead or redundant code and made the fixes you recommended. I also switched all the topics, materials, assesments, and LOs to insert in bulk for performance improvements.

laravel/resources/views/courses/wizard/header.blade.php
```diff
@@ -721,9 +721,8 @@ function campusOtherOption() {
         location.reload();
     });
 
-    $(document).ready(function () {
 	//This method is used to make sure that the proper amount of characters are entered so it doesn't exceed the max character limits
-    function validateMaxlength(e){
+    function validateMaxlength(){
```
FabianLCH: We should use the event object passed to this listener instead of the global even object (which is more fragile and thus less reliable)

python/services/syllabi_service/app/services/syllabus_parser.py
FabianLCH: Got it, all good then!

python/services/syllabi_service/app/services/syllabus_parser.py
FabianLCH: Noted! I feel like 12 rows is a safe sample number. If anything, we can change it in the future as needed

python/services/syllabi_service/app/services/syllabus_parser.py
```diff
@@ -1480,55 +1490,185 @@ def get_topics_from_table(doc: pymupdf.Document, topic_pages: list[int]) -> list
...
+    for row in rows:
+        for topic_column_index in topic_column_indexes:
+            if topic_column_index >= len(row):
+                continue
+
+            cell = row[topic_column_index].strip()
+            if not cell and header[topic_column_index] == "topic speaker" and topic_column_index > 0 and not header[topic_column_index - 1]:
+                cell = row[topic_column_index - 1].strip() # PyMuPDF can put Topic/Speaker text in the blank spacer column beside the header
+            if not cell and header[topic_column_index] in THEME_TOPIC_COLUMN_HEADINGS and topic_column_index > 0 and not header[topic_column_index - 1]:
+                cell = row[topic_column_index - 1].strip() # PyMuPDF can put values in the blank spacer column just before a topic-like header
```
FabianLCH: Since both conditionals check mostly the same things and are doing the same assignment, let's simplify this with:

`... and header[topic_column_index] in (THEME_TOPIC_COLUMN_HEADINGS | { "topic speaker" }) ...`

FabianLCH (changes requested):
Just need some very minor changes. After that, we'll be good to merge!

laravel/resources/views/courses/wizard/header.blade.php
NEOGDAWG: Done!

python/services/syllabi_service/app/services/syllabus_parser.py
NEOGDAWG: Done!

FabianLCH (approved):

PR #18, opened 2026-07-22
Course search functionality + saved filter presets

### LINKED ISSUE #13: Course content search across programs

NEOGDAWG:
This PR adds the main course search functionality for the curriculum mapping tool. It lets logged in users search across course content, narrow results with filters, view results by course or program, and save filter presets for later use.

- Added authenticated Course Search page at `/search`, with a new Course Search button in the main header navigation.
- Added database migrations for PostgreSQL generated `tsvector` search columns and GIN indexes so course content can be searched efficiently with good performance for larger datasets
- Implemented full-text search across course code, course number, course title, topics, learning outcomes, assessments, descriptions, materials, and program names.
- Added Course View and Program View, so users can either see matching courses directly or see results grouped by matching programs.
- Added property filters, course code filters, course level filters, and program filters to help narrow search results.
- Added saved filter preset support, including database storage, creating, applying, showing the current preset, and deleting saved filters.
- Added search result stats, result counts, pagination, highlighted snippets, and quick links to help make search results easier to review.
- Added safer highlighted search snippets so generated result previews can still highlight matches without rendering unsafe HTML injection.
- Added logic so direct program name matches still respect course code and course level filters when those filters are selected.
- Added a small info modal on the Course Search page explaining what the search does, query tips like `AND`, `OR`, quotes, and excluding terms, plus notes about saved filters and empty searches.
- Created a  `docs/search.md` file with the main implementation details, search behavior, routes, migrations, filters, saved presets, and testing notes.
- Wrote in-code documentation.

Notes:
- Empty searches are allowed, but they do not list every course or program by default.
- The search page requires users to be logged in and verified.
- Search ranking is based on where the query matched and how many useful matches were found across the course content.
- Saved filters are scoped to the logged-in user.
- Role access management has not been implemented yet for user - this will be the next step for this feature.

Tests:
- Added and updated feature tests for search behavior, filters, saved presets, program view behavior, snippet safety, and authenticated access.
- Use `php artisan test tests/Feature/SearchTest.php` to run the test suite for this feature.

**Please refer to `docs/search.md` for more details about the search implementation.**


docs/search.md
```diff
@@ -0,0 +1,224 @@
...
+
+Authenticated users can save named filter presets.
+
+Saved presets are stored in `saved_search_filters` with:
+
+- `user_id`
+- `name`
+- `filters` as JSONB
+
+The filters JSON stores the selected view, properties, course codes, course levels, and program IDs. Preset names are unique per user.
```
FabianLCH, line 165: Not super important, but it would be great if we include the `filters` object's schema or an example of what it may look like here

laravel/app/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,825 @@
...
+                $searchTerm,
+                $selectedProperties,
+                $selectedCourseCodes,
+                $selectedCourseLevels,
+                $selectedProgramIds,
+            );
+            $results = $resultsAndStats['results'];
+            $stats = $resultsAndStats['stats'];
+            $programMatches = $this->searchProgramNames($searchTerm, $selectedCourseCodes, $selectedCourseLevels);
+            $programResults = $this->groupCourseResultsByProgram($results,$programMatches,$selectedProgramIds);
```
FabianLCH: Is it necessary to do program search if `$selectedView` is `'courses'`? It looks like it's only used on the blade template if `$selectedView` is not `'courses'`. Maybe we can run the program search conditionally based on the selected view to save some effort

laravel/app/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,825 @@
...
+
+                    if ($minimum === 600) {
+                        $levelQuery->orWhere('courses.course_num', '>=', $minimum); //final level is open ended
+                    } else {
+                        $levelQuery->orWhereBetween(
+                            'courses.course_num',
+                            [$minimum, $minimum + 99] //in the query, normal levels become ranges (300 -> 300-399)
+                        );
+                    }
+                }
```
FabianLCH, line 292: We can simplify this down to the code in the `else` block since, even if the final level is technically open-ended, it should follow the same pattern (600-699)

laravel/app/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,825 @@
...
+     * @param string $searchTerm The normalized text to search for.
+     * @param array $courseCodes The selected course codes.
+     * @param array $courseLevels The selected course number levels.
+     * @param array $selectedProgramIds The selected program IDs used to restrict matching courses.
+     *
+     * @return Collection The matching courses with highlighted course snippets.
+     */
+    public function searchCourseNames(string $searchTerm, array $courseCodes, array $courseLevels, array $selectedProgramIds){
+        $searchText = "concat_ws(' ', courses.course_code, courses.course_num, courses.course_title)";
+        $normalizedSearchTerm = preg_replace('/^([A-Za-z]+)\s*(\d+)$/', '$1 $2', $searchTerm); //normalize course code/nums for better search
```
FabianLCH: The normalization here might not work as expected if someone includes at least part of the course title in the search term, or if they use the new notation for course codes (e.g. "FRST_V 100"). It may be better to search for 2 or more consecutive instances of spaces and replace them with a single space

laravel/app/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,825 @@
...
+
+    /**
+     * Applies the selected program filters from the search request, filtering only selected programs by building a sub query
+     * 
+     * @param Builder $query The query to be built on using SQL to apply program filters
+     * @param array $programIds User selected program Ids for filtering
+     * 
+     * @return Builder Query with slected program restrictions applied.
+     */
+    private function applyProgramFilters(Builder $query, array $programIds): Builder{
```
FabianLCH: Maybe we can rename this function to something like `filterByProgramIds` to make its general purpose clearer? The current name makes it sound like multiple filters can be applied, but the function only takes program IDs to filter by

laravel/app/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,825 @@
...
+
+    /**
+     * Calculates distinct course and program totals and match counts for each course property.
+     *
+     * @param Collection $matches Raw matches returned by all course property searches.
+     * @param Collection $results Combined course results with their associated programs.
+     *
+     * @return array The overall course, program, and property match totals.
+     */
+    public function calculateSearchStats(Collection $matches, Collection $results): array{
```
FabianLCH: To make what the arguments represent clearer, maybe we can rename them to:
- `$matches` > `$rawCourseMatches`
- `$results` > `$enrichedCourseResults`

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
+@extends('layouts.app')
+
+@section('content')
+    <style>
+        .search-page-header {
+            max-width: 780px;
+            margin: 0 auto 1.5rem;
+        }
```
FabianLCH, line 19: Suggesting some extra CSS to make the search controls align with the center of the main section a bit more nicely and leave some separation between controls and alerts:

```suggestion
        #app {
            display: flex;
            flex-direction: column;
        }

        .bg-body {
            flex-grow: 1;
            display: grid;
            align-items: center;
        }

        .search-page-header {
            max-width: 780px;
            margin: 1rem auto 1.5rem;
        }
```

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+            function renderCourseCodeOptions() {
+                const searchText = courseCodeSearch.value.trim().toLowerCase();
+                courseCodeOptionsContainer.innerHTML = '';
+
+                // Do not show course codes that are already selected.
+                const matchingCourseCodes = courseCodeOptions
+                    .filter(function (courseCode) {
+                        return !selectedCourseCodes.has(courseCode)
+                            && (!searchText || courseCode.toLowerCase().includes(searchText));
+                    });
```
FabianLCH: If too many course codes exist in the system, this would result in rendering a very long list of options which would potentially degrade performance. I'm thinking we can set a limit to the number of matches selected (e.g. 10) and change the filtering so it stops once those X matches have been found. If the user doesn't get the course code they are looking for, they can make their selection more specific

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+                const searchText = programSearch.value.trim().toLowerCase();
+                programOptionsContainer.innerHTML = '';
+
+                const matchingPrograms = programOptions
+                    .filter(function (program) {
+                        const programId = String(program.program_id);
+
+                        return !selectedPrograms.has(programId)
+                            && (!searchText || program.program.toLowerCase().includes(searchText));
+                    });
```
FabianLCH: Similar to course code selection, I think we can limit the program matches here to a specific amount to avoid rendering a really long list

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+                    addSavedFilterValue('program_ids[]', programId);
+                });
+            }
+
+            if (openSaveFilterModal && savedFilterValues) {
+                openSaveFilterModal.addEventListener('click', function () {
+                    prepareSavedFilterValues();
+                    closeSearchFilterMenu();
+                });
+            }
```
FabianLCH, line 1138: Not super important, but this makes it so `closeSearchFilterMenu` is called twice when the click event is triggered if saved filter values exist. You can simplify this to:
```suggestion
            if (openSaveFilterModal && savedFilterValues) {
                openSaveFilterModal.addEventListener('click', prepareSavedFilterValues);
            }
```
Since you're adding an event listener for `closeSearchFilterMenu` below

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+                    placeholder="Search courses"
+                    class="form-control search-input"
+                >
+
+                <button
+                    type="button"
+                    class="btn btn-outline-secondary search-action-button search-filter-button"
+                    id="searchFiltersButton"
+                    data-bs-toggle="dropdown"
+                    data-bs-auto-close="false"
```
FabianLCH: I think we can change this to keep the behaviour that closes the dropdown if the user clicks outside of it:
```suggestion
                    data-bs-auto-close="outside"
```

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+            const propertyOptions = Array.from(document.querySelectorAll('.property-filter-option'));
+
+            function updatePropertyControls() {
+                if (allProperties.checked) {
+                    propertyOptions.forEach(function (option) {
+                        option.checked = true;
+                        option.disabled = true;
+                    });
+                    return;
+                }
```
FabianLCH, line 963: Unchecking a 'select all' checkbox normally unchecks all 'individual' checkboxes, so I'm thinking reproducing that behaviour here might make using it more intuitive. If someone doesn't want to search by all properties, I can imagine usually their intention is to search for one or two specific properties, which would require manually unchecking all the other ones

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+
+            allProperties.addEventListener('change', updatePropertyControls);
+
+            propertyOptions.forEach(function (option) {
+                option.addEventListener('change', function () {
+                    allProperties.checked = propertyOptions.every(function (propertyOption) {
+                        return propertyOption.checked;
+                    });
+
+                    updatePropertyControls();
```
FabianLCH, line 984: Considering my comment on `updatePropertyControls`, I'm thinking it's worth separating the function's behaviour so the "All Properties" changing states operates independently from selecting all individual checkboxes. Currnetly, we would not be able to implement the suggestion in my comment as, given `updatePropertyControls` is called on every individual checkbox click, all individual checkboxes would get unchecked as long as at least one isn't checked already

laravel/resources/views/search/index.blade.php
```diff
@@ -0,0 +1,1275 @@
...
+
+        @if($programResults->hasPages())
+            <div class="mt-4 d-flex justify-content-center">
+                {{ $programResults->links() }}
+            </div>
+        @endif
+    @endif
+
+    <script>
+        document.addEventListener('DOMContentLoaded', function () {
```
FabianLCH, line 951: There's no code for it at the moment, but it might be a good idea to update the search summary chips whenever filters are updated or when the filter dropdown is closed (the latter would be more efficient as it would result in less DOM updates and the dropdown covers the summary chips anyway). The reason I bring this up is, there is no visual change outside of the dropdown menu when the user changes a filter until they perform a change, so they may think their filters did not apply

laravel/tests/Feature/SearchTest.php
FabianLCH, line 1: The tests are already super comprehensive, but I think we can add the following:

- User cannot apply filter owned by another user
- User cannot delete filter owned by another user

FabianLCH (changes requested):
Additions are looking great! My comments are mostly about improvements I think we should make before considering the PR done

docs/search.md
NEOGDAWG: Done!

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: Updated the normalization to support course codes within longer queries and underscored course-code notation. I've also added a test to ensure that new course code notations are supported.

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: done!

laravel/resources/views/search/index.blade.php
NEOGDAWG: That makes sense; I've added that behaviour.

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: Done!

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: Changed the names!

laravel/resources/views/search/index.blade.php
NEOGDAWG: I tried applying this suggestion, but openSaveFilterModal and openSavedFiltersModal are separate controls. Removing the explicit closeSearchFilterMenu() call caused the settings dropdown to remain open for Save Filter while closing for View Saved Filters, creating inconsistent behaviour. I also tried keeping the dropdown open while the modal was active, but that caused issues when closing or submitting the modal. Since the existing behaviour is consistent and the duplicate hide call is harmless, I kept the original implementation for now. Please let me know if you think we should still look into this simplification.

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: Good point. I updated this so direct program-name searching and result grouping only run when Program View is selected. Course View still displays and counts programs associated with matching courses, but it no longer performs the additional program-name search. I also updated the existing test to cover this behavior.

laravel/resources/views/search/index.blade.php
NEOGDAWG: Good Idea, I've added a filter options limit of 10 to both the cases you mentioned!

laravel/resources/views/search/index.blade.php
NEOGDAWG: Addressed with the earlier All Properties update. Only changing All Properties now checks or unchecks every individual option, while individual changes only recalculate the All Properties state. This preserves partial selections.

laravel/resources/views/search/index.blade.php
NEOGDAWG: Good catch - I overlooked this. Added this so the summary chips update when the settings dropdown finishes closing!

laravel/tests/Feature/SearchTest.php
NEOGDAWG: Added!

laravel/resources/views/search/index.blade.php
FabianLCH: My bad! I didn't realize the slight difference between the two variables and thought both referred to the same modal. We can keep this as is

FabianLCH (approved):
Thanks for making the changes, Kabir! This PR is good to go

PR #19, opened 2026-07-27
Course material file upload, content indexing and topic extraction

### LINKED ISSUE #14: Course learning material upload for topic coverage analysis
# Description

During the curriculum development process, often, there are gaps in instructors and program designers' understanding of topic coverage across courses in the same program or programs belonging to the same faculty. To bridge this gap, the Curriculum Development Tool should support uploading and extracting key information from course materials, which can then be integrated into system search functionality, reporting, etc.

# Tasks
1. Identify the type and format of learning materials that should be supported (e.g. lecture slides, reports, papers, etc.)
2. Define categories of information to extract from learning materials (e.g. topics, technical keywords, etc.)
3. Test multiple methods (LLMs, topic extraction model, etc.) to extract topics from content
4. Design and build app functionality to allow upload of learning materials
5. Expand app and supporting micro-service functionality to support course material upload, analysis and extraction

# Requirements 
- Support the upload of learning materials as PDFs
- Set a strict limit for uploaded materials (which will later be tuned as needed)
- Extract material context as text directly or via OCR if text available in images
- Index material content in the database for search purposes
- Extract course topics from materials
- Integrate into the "Course Materials" step of the course creation wizard

# Additional Context

## Expected types of learning materials
- Articles, which may contain:
  - Graphs/tables
  - Images (tree and plant species, etc.)
- Images with short descriptions
- Lecture slides (originally saved as PowerPoint slides)
- Video transcripts
- Podcast transcripts
- Textbook modules

Apps247:
This PR primarily adds a new microservice for text and topic extraction from PDFs associated with course materials. It also allows full text search of these course material PDFs.

**Additions**
- Text and topic extraction microservice
   - supports OCR for scanned PDFs
   - has special topic suggestion handling for Slides & Articles using BERTopic and font properties
- Frontend UI for uploading course material PDFs
- DB text indexing for uploaded PDFs using `tsvector`s.
- Temporary Search UI with text highlighting and thumbnails (see note)
- Frontend & DB changes to support associating one or more files to 
- Frontend & DB changes to support adding/removing topics, requesting topic suggestions, and reviewing these suggestions
- Python tests for various handlers and postprocessing patterns
- Laravel tests for Suggested Topic review functionality
- End to End Test (currently not working, see last note)

**Notes**:
- The search UI is marked (Temp), and can be accessed via a course page or program page. The logic behind the search is the same as that of  #18, so it should be easy to integrate it into the main search functionality once that is merged.
- The NLP models used are currently pulled (and cached) from HuggingFace by their Python libraries during runtime. We may want to have local copies of these models that our code can use.
- This PR also refactors some of the `composer test` set up to share common fixtures across tests
- The E2E test currently fails, because of a known bug with Pest for file uploads. Someone made a fix and it should be merged in to the new version soon. Please ignore `TopicReviewTest.php` until then.


docs/FastAPITextExtractionService.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+### `POST /extract`
+
+The main endpoint. Accepts an [`ExtractRequest`](../python/services/text_extraction_service/app/schemas.py) with:
+
+- `file` (string, required): base64-encoded PDF content
+- `ocr_enabled` (boolean, default `false`): whether to apply OCR on low-text pages
+- `extraction_engine` (string, default `"tesseract"`): `"tesseract"` or `"textract"`
+- `ocr_threshold` (integer, default `0`): page text length below which OCR is triggered
```
FabianLCH: We should specify here whether this refers to character length, line length, etc. Should also specify what happens if the value is set to 0 (OCR triggered regardless? OCR never triggered?)

docs/FastAPITextExtractionService.md
```diff
@@ -0,0 +1,156 @@
...
+
+### BERTopic Extraction
+
+The BERTopic extractor in [`bertopic_extractor.py`](../python/services/text_extraction_service/app/services/topic_extraction/bertopic_extractor.py):
+
+1. Lemmatizes text with spaCy to collapse plurals
+2. Splits text into overlapping windows
+3. Fits BERTopic with the `all-mpnet-base-v2` embedding model.
+4. Extracts top keywords from each cluster
+5. Returns up to `TOPICS_COUNT` (currently 100) topics, then sorts by relevance. This limit should ideally never be reached, and is only there to prevent the database exploding with topics. We can increase the limit any time.
```
FabianLCH: I'm aware of the note about the topic limit ideally never being reached, but I think the default limit should be way lower. I don't think anyone would realistically want to extract more than 20 topics from a single material

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+            'file' => ['required', 'file', 'mimes:pdf', 'max:51200'],
+            'ocr_enabled' => ['sometimes', 'boolean'],
+            'extraction_engine' => ['required_if:ocr_enabled,1', 'in:tesseract,textract'],
+            'ocr_threshold' => ['sometimes', 'integer', 'min:0', 'max:100000'],
+        ]);
+
+        $uploaded = $request->file('file');
+        $diskPath = 'course-materials/' . $course_id . '/' . Str::uuid()->toString() . '.pdf';
+
+        Storage::disk('local')->put($diskPath, file_get_contents($uploaded->getRealPath()));
```
FabianLCH: Maybe we can use `$uploaded->storeAs(...)` here to avoid having to explicitly read the contents of the temp uploaded file (may also have better performance depending on how it creates the persisted file)

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        SuggestedTopic::where('course_material_file_id', $file->course_material_file_id)
+            ->where('status', SuggestedTopic::STATUS_PENDING)
+            ->whereIn(DB::raw('LOWER(topic)'), $lowerKeys)
+            ->delete();
+    }
+
+
+    private function assertIsEditor(int $course_id): void
```
FabianLCH: Not super important, but I think we can rename function to `verifyUserIsEditor`. The `assert...` naming convention is usually reserved for tests, so it's worth keeping the distinction via naming

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+
+        $file = CourseMaterialFile::where('course_material_file_id', $file_id)
+            ->where('course_material_id', $material_id)
+            ->firstOrFail();
+
+        $request->validate([
+            'decisions' => 'required|array|min:1',
+            'decisions.*.id' => 'required|integer|exists:suggested_topics,suggested_topic_id',
+            'decisions.*.action' => 'required|in:confirm,reject',
+        ]);
```
FabianLCH, line 223: Let's move this above the `$file` definition above to avoid the DB query if the request fails validation

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+
+        $decisions = $request->input('decisions');
+        $confirmIds = array_column(array_filter($decisions, fn($d) => $d['action'] === 'confirm'), 'id');
+        $rejectIds  = array_column(array_filter($decisions, fn($d) => $d['action'] === 'reject'), 'id');
+
+        DB::transaction(function () use ($file, $confirmIds, $rejectIds) {
+            if ($confirmIds) {
+                $topicTexts = SuggestedTopic::where('course_material_file_id', $file->course_material_file_id)
+                    ->where('suggested_topic_id', $confirmIds)
```
FabianLCH: I think here is should be:
```suggestion
                    ->whereIn('suggested_topic_id', $confirmIds)
```

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+                    ->where('status', SuggestedTopic::STATUS_PENDING)
+                    ->pluck('topic')
+                    ->all();
+
+                $this->applyTopicDecisions($file, $topicTexts);
+            }
+
+            if ($rejectIds) {
+                SuggestedTopic::where('course_material_file_id', $file->course_material_file_id)
+                    ->where('suggested_topic_id', $rejectIds)
```
FabianLCH: Similarly:
```suggestion
                    ->whereIn('suggested_topic_id', $rejectIds)
```

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+                    ->where('status', SuggestedTopic::STATUS_PENDING)
+                    ->update(['status' => SuggestedTopic::STATUS_REJECTED]);
+            }
+        });
+
+        return redirect()->back()->with('success', 'Topics reviewed.');
```
FabianLCH: As an extra piece of helpful info, let's include the number of topics accepted vs rejected in the success message here

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+                $newNames
+            );
+
+            DB::table('course_topics')->insert($rows);
+
+            $newLowerNames = array_map('strtolower', $newNames);
+            $newIds = CourseTopic::where('course_id', $file->courseMaterial->course_id)
+                ->whereIn(DB::raw('LOWER(topic)'), $newLowerNames)
+                ->pluck('course_topic_id')
+                ->all();
```
FabianLCH, line 345: We should be able to shorten this and skip another query by using `insertGetId` assuming the `course_topics` table's ID is auto-incrementing:
```suggestion
            $newIds = DB::table('course_topics')->insertGetId($rows, 'course_topic_id');
```

laravel/app/Http/Controllers/CourseMaterialFileController.php
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        $allTopicIds = array_unique(array_merge($existingIds, $newIds));
+
+        if ($allTopicIds) {
+            $file->topics()->syncWithoutDetaching($allTopicIds);
+        }
```
FabianLCH: `$newIds` shouldn't overlap with `$existingIds`, so you should be able to just call:

```suggestion
        $file->topics()->syncWithoutDetaching($newIds);
```
`syncWithoutDetaching` keeps existing relationships, so you just need to worry about the new IDs. This probably also means you don't have to worry about storing existing IDs above.

laravel/app/Jobs/IndexCourseMaterial.php
```diff
@@ -0,0 +1,179 @@
...
+                        'material_type' => $file->courseMaterial?->type,
+                        'existing_topics' => $existingTopics,
+                    ]);
+            } else {
+                $absolutePath = Storage::disk('local')->path($file->file_path);
+                $fileBytes = file_get_contents($absolutePath);
+
+                $response = Http::timeout($this->timeout)
+                    ->post(config('services.text_extraction.base_url') . '/extract', [
+                        'file' => base64_encode($fileBytes),
```
FabianLCH: Given how encoding data in base64 format tends to considerably increase size, I'm thinking it might be best to send the file as a multi-part request and have FastAPI process the it as an `UploadedFile`. This should especially work better for large files

laravel/app/Jobs/IndexCourseMaterial.php
```diff
@@ -0,0 +1,179 @@
...
+            Log::error("IndexCourseMaterial failed for file {$file->course_material_file_id}: " . $exception->getMessage());
+            $file->update([
+                'status' => CourseMaterialFile::STATUS_FAILED,
+                'error_message' => mb_substr($exception->getMessage(), 0, 1000),
+            ]);
+            throw $exception;
+        }
+    }
+
+    private function chunkDBRow(CourseMaterialFile $file, int $pageNumber, string $text): array
```
FabianLCH: Maybe we can rename this function to `pageAsDBChunk` to make its purpose clearer at first glance

laravel/app/Jobs/IndexCourseMaterial.php
```diff
@@ -0,0 +1,179 @@
...
+        $existingTopics = $file->suggestedTopics()
+            ->pluck('topic')
+            ->flip();
+
+        $rows = [];
+        foreach ($topics as $topic) {
+            $text = trim($topic['topic'] ?? '');
+            if ($text === '' || $rejectedTopics->has($text) || $existingTopics->has($text)) {
+                continue;
+            }
```
FabianLCH: `$rejectedTopics` is a subset of `$existingTopics` so you can just get `$existingTopics` and check if the topic to insert is in that array

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -1,308 +1,680 @@
...
-                                            <div class="invalid-tooltip">
-                                                Please provide a material name.
+                                <div class="modal-body">
+                                    <form id="addCourseMaterialsForm" class="needs-validation" novalidate>
+                                        <!-- Materials have multiple fields, so the add form collects one complete material row. -->
+                                        <div class="row justify-content-between align-items-end m-2">
+                                            <div class="col-4">
+                                                <label for="courseMaterialName" class="form-label fs-6"><b>Name</b></label>
+                                                <input id="courseMaterialName" class="form-control"
+                                                    oninput="validateMaxlength(event)" onpaste="validateMaxlength(event)"
```
FabianLCH: Minor tip here, you don't have to call the event listener with the global event object explicitly. You can just set `oninput` to the function name and it will automatically be called with the appropriate event object

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -1,308 +1,680 @@
...
+                                                    <div id="ocrAdvBody" class="accordion-collapse collapse"
+                                                        data-bs-parent="#ocrAdvAccordion">
+                                                        <div class="accordion-body p-2 row">
+                                                            <div class="col-md-6">
+                                                                <label class="form-label small mb-0">Extraction Engine</label>
+                                                                <select name="extraction_engine"
+                                                                    class="form-select form-select-sm"
+                                                                    onchange="toggleOcrThresholdSetting(this)">
+                                                                    <option value="tesseract">Tesseract</option>
+                                                                    <option value="textract">AWS Textract (cloud)</option>
```
FabianLCH: Let's add a message here (either always shown or conditionally shown) explaining data will be temporarily sent to cloud servers if the AWS Textract option is chosen

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -1,308 +1,680 @@
...
+                                                                </select>
+                                                            </div>
+                                                            <div id="ocrThresholdSetting" class="col-md-6">
+                                                                <label class="form-label small mb-0">OCR threshold (characters)</label>
+                                                                <input type="number" name="ocr_threshold" min="0"
+                                                                    max="100000" value="0"
+                                                                    class="form-control form-control-sm">
```
FabianLCH: Let's add some help text below explaining pages in the uploaded document with less characters than the threshold will be processed using OCR (maybe without the technical term)

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -1,308 +1,680 @@
...
+                                                                        <th class="text-center">Status</th>
+                                                                        <th class="text-center">Pages</th>
+                                                                        <th class="text-center">Actions</th>
+                                                                    </tr>
+                                                                </thead>
+                                                                <tbody>
+                                                                    @foreach($courseMaterial->files as $file)
+                                                                                                                    <tr>
+                                                                                                                        <td class="align-middle">
+                                                                                                                            <a
```
FabianLCH: The indenting in this section is a bit wild 😆

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -1,308 +1,680 @@
...
+                { cls: "material-name",     field: "name" },
+                // { cls: "material-type",     field: "type" },
+                { cls: "material-desc",     field: "description" },
+                { cls: "material-url",      field: "url" },
+                { cls: "material-required", field: "is_required" },
+            ];
+
+            // TODO: Should we just move materialTypes to JS here?
+            //       Kept in PHP for now in case we move it to DB later
+            const materialTypes = @json($materialTypes);
```
FabianLCH: We can keep it in PHP as it's also used to render some HTML above!

laravel/resources/views/courses/wizard/step10.blade.php
```diff
@@ -1,308 +1,680 @@
...
+                feedback.textContent = '';
+                feedback.classList.add('d-none');
+            }
+        }
+
+        function addCourseMaterial() {
+            materials.push({
+                id: null,
+                name: $('#courseMaterialName').val(),
+                // TODO: What should we do if they leave the input blank after choosing 'Other'? Just store blank?
```
FabianLCH: Users who choose "Other" should be required to provide a type (i.e. cannot be blank)

laravel/resources/views/courses/coverageAnalysis.blade.php
```diff
@@ -0,0 +1,64 @@
...
+                        <a href="{{ route('course.material.files.view', ['course' => $course->course_id, 'material' => $result->course_material_id, 'file' => $result->file_id]) }}#page={{ $result->page_number }}"
+                           target="_blank" class="text-decoration-none text-reset">
+                            <div class="border rounded p-3 mb-2">
+                                <div class="row g-3 align-items-center">
+                                    <div class="col-12 col-md">
+                                        <div class="mb-1">
+                                            <strong>{{ $result->file_name }}</strong>
+                                            <span class="text-muted ms-2">Page {{ $result->page_number }}</span>
+                                        </div>
+                                        <div>{!! $result->snippet !!}</div>
```
FabianLCH: Since the snippet comes from a user-uploaded document, we should not print it unescaped. A malicious user could upload a document with malicious HTML which could then be evaluated by someone else with access (e.g. an admin) who is searching the course.

Currently, we mark our snippet selection using `<mark>` when making the DB query. So, what I think we can do is:
1. Use a special character sequence to mark the selection start and end
2. Split the snippet into sections based on the special characters
3. Display each section with escaping, explicitly wrapping the selection/match section with `<mark>` tags

An alternative would be to:
1. Use alternative selection indicators like `__startsel__` and `__endsel__` (doesn't have to be this exactly, as long as the indicators are distinguishable, unique enough to not overlap with actual file content and do not use special characters that would get escaped)
2. Call Laravel's `e()` function on `$result->snippet` to escape special characters
3. Replace `__startsel__` and `__endsel__` with `<mark>` and `</mark>` respectively
4. Output the result with the `{!! ... !!}` operator

The second approach might be more feasible if there are multiple fragments. The first approach is still doable, though we would have to make assumptions on the maximum number of segments of split text based on what we set for MaxFragments.

laravel/resources/views/courses/material_file.blade.php
```diff
@@ -0,0 +1,560 @@
...
+        @endforeach
+    ];
+
+    let editMode = false;
+
+    let topics = [
+        @foreach ($file->topics as $topic)
+            { id: {{ $topic->course_topic_id }} },
+        @endforeach
+    ];
```
FabianLCH: Is there a need to format `topics` as a list of objects with one property? It seems this could just be a flat list of IDs (would also remove the need to call `.map()` below)

laravel/resources/views/courses/material_file.blade.php
```diff
@@ -0,0 +1,560 @@
...
+
+<script>
+document.addEventListener('DOMContentLoaded', () => {
+    document.querySelectorAll('[data-bs-toggle="tooltip"]').forEach(el => new bootstrap.Tooltip(el));
+
+    const tbody = document.getElementById('topics-tbody');
+    const editBtn = document.getElementById('edit-topics-btn');
+    const saveBtn = document.getElementById('save-topics-btn');
+    const cancelBtn = document.getElementById('cancel-topics-btn');
+    const addBtn = document.getElementById('add-topic-btn');
```
FabianLCH: The variables here could use clearer names, e.g. `tbody` > `tbodyFileTopics` (there are multiple tables plus three different types of topics: 'course', 'file' and 'suggested'), `editBtn` > `editTopicsBtn`, etc.

laravel/resources/views/courses/material_file.blade.php
FabianLCH: Could also rename this to something like `fileTopics` or `fileTopicIds` to make what it represents clearer

laravel/resources/views/programs/coverageAnalysis.blade.php
```diff
@@ -0,0 +1,64 @@
...
+                        <a href="{{ route('course.material.files.view', ['course' => $result->course_id, 'material' => $result->course_material_id, 'file' => $result->file_id]) }}#page={{ $result->page_number }}"
+                           target="_blank" class="text-decoration-none text-reset">
+                            <div class="border rounded p-3 mb-2">
+                                <div class="row g-3 align-items-center">
+                                    <div class="col-12 col-md">
+                                        <div class="mb-1">
+                                            <strong>{{ $result->file_name }}</strong>
+                                            <span class="text-muted ms-2">Page {{ $result->page_number }}</span>
+                                        </div>
+                                        <div>{!! $result->snippet !!}</div>
```
FabianLCH: Similar to the comment on `courses/wizard/coverageAnalysis.blade.php`, we should escape the snippet before outputting to HTML

laravel/tests/Feature/SuggestedTopicTest.php
```diff
@@ -0,0 +1,496 @@
...
+use App\Models\SuggestedTopic;
+use App\Models\User;
+use Carbon\Carbon;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\DB;
+use Tests\TestCase;
+
+class SuggestedTopicTest extends TestCase
+{
+    public function test_suggested_topic_confirmation(): void
```
FabianLCH, line 19: We should also check here that the suggested topic is deleted from the DB when the user confirms it

python/services/text_extraction_service/app/api/routes.py
```diff
@@ -0,0 +1,90 @@
...
+    title="Course Material Text & Topic Extraction Service",
+    version="0.1.0",
+    docs_url="/docs",
+    redoc_url="/redoc",
+)
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=settings.ALLOWED_ORIGINS,
+    allow_credentials=True,
```
FabianLCH: Since we're using the `*` wildcard for `allow_methods` and `allow_headers`, this should probably be set to `False`

python/services/text_extraction_service/app/api/routes.py
```diff
@@ -0,0 +1,90 @@
...
+)
+
+
+@app.get("/health")
+async def health_check() -> dict[str, str]:
+    return {"status": "ok"}
+
+
+@app.post("/extract", response_model=ExtractResponse)
+def extract(request: ExtractRequest) -> ExtractResponse:
```
FabianLCH: If `request.extraction_engine` is set to `textract` and AWS credentials are not configured (empty or None), let's make the endpoint immediately return HTTP 501

laravel/app/Jobs/IndexCourseMaterial.php
```diff
@@ -0,0 +1,179 @@
...
+                    ->get()
+                    ->map(fn($chunk) => ['page_number' => $chunk->page_number, 'content' => $chunk->content])
+                    ->all();
+
+                $response = Http::timeout($this->timeout)
+                    ->post(config('services.text_extraction.base_url') . '/refresh-topics', [
+                        'pages' => $pages,
+                        'material_type' => $file->courseMaterial?->type,
+                        'existing_topics' => $existingTopics,
+                    ]);
```
FabianLCH, line 68: To avoid overloading the service with very large files, I'm thinking we should maybe split this into multiple requests with a maximum number of pages processed at a time. Say, we only ever do at most 10 pages at a time, and results are concatenated together as they are received

python/services/text_extraction_service/app/services/text_readers/textract_extractor.py
```diff
@@ -0,0 +1,128 @@
...
+                logger.info("Synchronous Textract failed: %s; falling back to async", e)
+        else:
+            return self._extract_async(session, file_bytes), page_count
+
+    def _extract_sync(self, session: boto3.Session, file_bytes: bytes) -> list[ExtractedPage]:
+        """Single-page extraction using the synchronous detect_document_text API."""
+        response = session.client("textract").detect_document_text(Document={"Bytes": file_bytes})
+        return _parse_textract_response(response)
+
+    def _extract_async(self, session: boto3.Session, file_bytes: bytes) -> list[ExtractedPage]:
```
FabianLCH: Since this method isn't actually asynchronous, I'm thinking a better name would be `_run_extract_job`

python/services/text_extraction_service/app/services/text_readers/textract_extractor.py
```diff
@@ -0,0 +1,128 @@
...
+
+        s3_key = f"textract-jobs/{int(time.time())}-{os.urandom(4).hex()}.pdf"
+        s3.put_object(Bucket=self.settings.AWS_S3_BUCKET, Key=s3_key, Body=file_bytes)
+
+        job_id = textract.start_document_text_detection(
+            DocumentLocation={"S3Object": {"Bucket": self.settings.AWS_S3_BUCKET, "Name": s3_key}}
+        )["JobId"]
+        logger.info("Started async Textract job %s", job_id)
+
+        return self._poll_job(textract, job_id)
```
FabianLCH: Currently it looks like there is no cleanup for files temporarily uploaded to S3. Maybe we can use a `try...except...finally` block here to ensure S3 temp files are deleted after the job completes (successfully or otherwise)

python/services/text_extraction_service/app/services/topic_extraction/faiss_playground.py
FabianLCH, line 1: I'm guessing this was temporarily created to explore topic extraction using FAISS, right? Do we still need it and the `faiss` library?

python/services/text_extraction_service/app/services/topic_extraction/postprocessor.py
```diff
@@ -0,0 +1,90 @@
...
+        topic_lower = t.topic.lower()
+        if topic_lower not in seen:
+            seen.add(topic_lower)
+            deduped_topics.append(t)
+            
+    # Remove topics with numbers in them
+    number_filtered_topics = []
+    for t in deduped_topics:
+        # TODO: Should this be not all instead of not any? Can any topics actually have numbers?
+        if not any(char.isdigit() for char in t.topic):
```
FabianLCH: This should probably be:
```suggestion
        # TODO: Should this be not all instead of not any? Can any topics actually have numbers?
        if not all(char.isdigit() or char.isspace() for char in t.topic):
```
since topics can have _some_ digits but shouldn't be **all** digits (and space-separated digits only still counts as 'all digits')

python/services/text_extraction_service/app/services/topic_extraction/type_specific_handlers.py
```diff
@@ -0,0 +1,181 @@
...
+
+from app.schemas import ExtractedLine, ExtractedPage, Topic
+from app.services.topic_extraction import postprocessor
+from app.services.topic_extraction import bertopic_extractor as extractor
+
+
+class MaterialTypeHandler:
+    """Base handler - extracts keywords from text only"""
+
+    def match_topics(self, text: str, topics: list[str], min_count: int = 1) -> list[Topic]:
```
FabianLCH, line 11: Suggesting a docstring here to make the method's purpose clearer:
```suggestion
    def match_topics(self, text: str, topics: list[str], min_count: int = 1) -> list[Topic]:
        """ Filters `topics` for items that appear at least `min_count` times within `text` and returns them. Matching is case-insensitive. """
```

python/services/text_extraction_service/app/services/topic_extraction/type_specific_handlers.py
```diff
@@ -0,0 +1,181 @@
...
+        
+        matched_topics = self.match_topics(preprocessed_text, existing_topics)
+
+        return matched_topics
+
+    def refresh_topics(self, pages: list[ExtractedPage], existing_topics: list[str] = []) -> list[Topic]:
+        """Extract match-based and keyword-based topics without font-based topics.
+        For the default Material handler, this is the same because only matched topics are used"""
+        # TODO: Check with Fabian if we should run BERTopic keyword extraction on default material types too
+        return self.extract_topics(pages, existing_topics)
```
FabianLCH: I don't think it's necessary to run BERTopic on the base handler. I'd rather we only run BERTopic explicitly in type-specific handlers once we know we will benefit from it

python/services/text_extraction_service/app/services/topic_extraction/type_specific_handlers.py
```diff
@@ -0,0 +1,181 @@
...
+
+def _to_topics(texts, source) -> list[Topic]:
```
FabianLCH: Not super important, but I think we can have this as a method of `MaterialTypeHandler`. The type-specific handlers are subclasses of `MaterialTypeHandler`, so those can call it via `self._to_topics`. Also, let's add type hints to the arguments!

python/services/text_extraction_service/tests/test_handlers.py
```diff
@@ -0,0 +1,170 @@
...
+        assert not any(t.source == "font" for t in topics_refreshed)
+        topics_extracted = handler.extract_topics([page], existing_topics=["forestry"])
+        print("Extracted topics:")
+        print([t.topic for t in topics_extracted])
+        assert any(t.topic == "Title slide" for t in topics_extracted)
+        assert any(t.topic == "forestry" for t in topics_extracted)
+
+
+class TestArticleHandler:
+    def test_heading_detection_and_topics(self):
```
FabianLCH: For better organization, I think this test can be split into at least 3 separate tests

python/services/text_extraction_service/tests/test_handlers.py
```diff
@@ -0,0 +1,170 @@
...
+        assert len(topics) == 1 and topics[0].topic == "Climate change"
+
+        page2 = _make_page(["Forest ecology"])
+        topics2 = handler.refresh_topics([page2], existing_topics=["Forest"])
+        assert len(topics2) == 1 and topics2[0].source == "match"
+
+
+
+class TestSlidesHandler:
+    def test_title_topics(self):
```
FabianLCH: This test can also be split into separate tests

FabianLCH (changes requested):
Haven't had a chance to set things up locally and manually try out the changes yet, but I wanted to at least provide a review based off the code since we're short on time. No major changes needed for now, but please have a look at my comments!

laravel/resources/views/courses/wizard/step10.blade.php
Apps247: Added this "If a page has fewer typed characters than the threshold, then it will be scanned as an image. Set a higher threshold if you have pages that combine text and figures."

Wasn't sure how best to explain to users, please lmk if it should be improved

python/services/text_extraction_service/app/services/text_readers/textract_extractor.py
Apps247: When the S3 bucket is created in `deploy_script.py`, we set a lifecycle of 1 day for all items in `textract-jobs/`.
As long as we create it with the deploy script, this shouldn't be an issue. Do you recommend doing the immediate cleanup anyway?

python/services/text_extraction_service/app/services/topic_extraction/faiss_playground.py
Apps247: Sorry, didn't mean to commit this one. This was my original plan for the `match_topics` method. Using the `faiss-cpu` package is not as heavy as BERTopic, but will match topics with slight variations even if they aren't exact strings. The examples I tried in the playground ran almost instantly. Didn't get around to implement this due to time, will remove this file.

Noting this here for future improvers if any!

python/services/text_extraction_service/app/api/routes.py
Apps247: This setup is also used in the LO Mapping service and the Syllabi service - should I set it to `False` in those as well?

laravel/app/Jobs/IndexCourseMaterial.php
Apps247: BERTopic works better the longer the source text, so for slides and articles, it'd be losing out on context. Especially slides, since there aren't too many words per page.
For `/refresh-topics`, only the pages' text is sent, not the actual PDF - so it wouldn't be as large as the original file.
Maybe we break it up by character count or word count instead, to ensure the NLP model has enough to work with per request?

laravel/app/Http/Controllers/CourseMaterialFileController.php
Apps247: We can accept multiple suggested topics before saving, and it seems like `insertGetId` doesn't work with multiple different ids. I just searched up on it now, please correct me if not!

laravel/app/Http/Controllers/CourseMaterialFileController.php
Apps247: I removed the unnecessary if guard. But `$newIds + $existingIds` was used because a topic suggested by BERTopic may already exist in the course's topics. If such a topic suggestion is accepted, it should sync to the already existing id of that `CourseTopic`, rather than being synced to a newly created `CourseTopic`.

python/services/text_extraction_service/app/api/routes.py
FabianLCH: Oh yes, it should be `False` on those too

laravel/app/Jobs/IndexCourseMaterial.php
FabianLCH: Ah, noted! We can keep it as is for the time being then. We can explore it more later, but I feel like breaking by character count will take lots of trial and error to find the right value unless there's some minimum/ideal value for BERTopic

laravel/app/Http/Controllers/CourseMaterialFileController.php
FabianLCH: Woops, you are right. Looks like `insertGetId` only returns the ID of the last inserted item, so it wouldn't help us here. Disregard the suggestion!

laravel/app/Http/Controllers/CourseMaterialFileController.php
FabianLCH: Ah, noted! You are right; both need to be included here otherwise the intermediate table would be missing associations between existing topics and the material file

laravel/resources/views/courses/coverageAnalysis.blade.php
Apps247: I went with the second approach you recommended. I also refactored out the repeated logic and UI to a `SearchedTextHighlighter.php` and a `search-result.blade.php`, to reduce repetition and potentially benefit the main search too

python/services/text_extraction_service/app/services/text_readers/textract_extractor.py
Apps247: Added the immediate cleanup, it's better since it can prevent clogging just before the term starts

NEOGDAWG:
I have finished integrating the main Course Search functionality into this branch and addressed the remaining cleanup items.

- Added indexed course material PDF content as a searchable property and filter.
- Added material-content ranking, highlighted snippets, PDF filenames, and matching page links.
- Limited material-content search to files with an `INDEXED` status.
- Added tests and updated the search documentation.
- Removed the temporary course-material search UI and its unused routes/controller.
- Completed and corrected the course/program guide modal sections.
- Fixed course navigation so course links consistently open the Course Description step (the first step in course creation) instead of the CLO section.
- Added stricter course-material file ownership validation to avoid validation issues.
- Minor clean ups and refactors

I also completed final manual testing of the material-content search using multiple indexed PDFs and confirmed the expected courses, snippets, filenames, and page numbers appeared. The Laravel test suite passes. This should now be ready for re-review. Thank you!

laravel/app/Jobs/IndexCourseMaterial.php
```diff
@@ -68,16 +68,18 @@ public function handle(): void
...
+                $metadata = json_encode([
+                    'ocr_enabled' => (bool) $file->ocr_enabled,
+                    'extraction_engine' => $file->extraction_engine,
+                    'ocr_threshold' => (int) $file->ocr_threshold,
+                    'material_type' => $file->courseMaterial?->type,
+                    'existing_topics' => $existingTopics,
+                ]);
 
                 $response = Http::timeout($this->timeout)
+                    ->attach('file', fopen($absolutePath, 'r'), basename($absolutePath))
```
FabianLCH: To avoid leaving an open file pointer while the job executes (and possibly beyond that), I think we can use `Storage::get($file->file_path)` to get the file contents and `$file->file_name`. This should also remove the need to define `$absolutePath` above

python/services/text_extraction_service/app/api/routes.py
```diff
@@ -39,18 +37,34 @@ async def health_check() -> dict[str, str]:
...
+    except Exception as e:
+        raise HTTPException(
+            status_code=422,
+            detail=f"Invalid metadata JSON format: {e}",
+        )
+
+    if request_metadata.extraction_engine == "textract" and not (settings.AWS_ACCESS_KEY_ID and settings.AWS_SECRET_ACCESS_KEY):
+        raise HTTPException(
+            status_code=501,
+            detail="Textract extraction requires AWS credentials; configure AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.",
```
FabianLCH: Not super important, but it might be best to make this message a bit more generic. For example, something like:

"Textract extraction is not configured."

Then, we can leave the full details for the logs

FabianLCH (changes requested):
PR is almost good to go! Other than the two review comments, the only other change I think we should make is compress the guide images a little to reduce their size using something like squoosh

python/services/text_extraction_service/app/api/routes.py
NEOGDAWG: Done! The missing AWS credential details are recorded using the existing service logger now also.

laravel/app/Jobs/IndexCourseMaterial.php
NEOGDAWG: Done!

NEOGDAWG:
> PR is almost good to go! Other than the two review comments, the only other change I think we should make is compress the guide images a little to reduce their size using something like squoosh

Sounds good! I compressed the guide images and updated their references. This reduced their total size by about 91% while keeping them clear and readable.

FabianLCH (approved):
PR is ready for merging!

Issue #20, opened 2026-08-05
Track remaining role access management edge cases


NEOGDAWG:
## Summary

Course Search now uses the same materialized access tables as the dashboard and course authorization. Direct access comes from `course_users`, elevated role access comes from `course_user_role`, and administrators bypass search restrictions.

During implementation and testing, several remaining edge cases were identified. They are not blockers for the current search feature but should be revisited as role and organization management develops.

## Organization ownership uses text matching

The `courses` and `programs` tables store campus, faculty, and department as text. The application matches these names against `campuses`, `faculties`, and `departments` when creating `course_user_role` and `program_user_role` rows.

This requires exact names. For example, `Forest Resources Management` would not match `Department of Forest Resources Management`, potentially preventing the expected access rows from being created.

This is acceptable for now because organizations are added through seeders. A future migration could associate existing courses and programs with faculty and department IDs, then use those IDs for access management.

## Faculty-wide access flags cannot be updated directly

Faculty-wide access is stored in:

- `department_head.has_access_to_all_courses_in_faculty`
- `program_user_role.has_access_to_all_courses_in_faculty`

Initial role assignment creates the required `course_user_role` rows. However, there is no supported flow for updating this flag later.

Changing it directly from false to true does not create missing rows, while changing it from true to false does not remove old rows. The current workaround is removing and reassigning the role.

A future fix could add support to update the flag and rebuild the affected access rows without removing access granted through another source.

## Faculty course-code changes do not synchronize access

Faculty course-code ownership is stored in `faculty_course_codes` and initialized through `FacultyCourseCodeSeeder`.

Adding a course code to a faculty does not create `course_user_role` rows for existing courses with that code. As a result, users with faculty-wide access will not automatically gain access to those existing courses. Removing or changing a mapping can similarly leave outdated access rows behind.

A future synchronization command, service, or data migration should update the mapping and rebuild the affected access rows.

## Shared course codes need a defined ownership rule

The `faculty_course_codes` table allows the same course code to be associated with multiple faculties. However, some course creation, update, duplication, and syllabus import paths use the first matching faculty.

If multiple mappings exist, faculty-wide `course_user_role` rows may only be created for one faculty, causing inconsistent access.

## Materialized access should be auditable

Search, dashboard, and page authorization depend on `course_user_role` and `program_user_role` remaining synchronized.

A future reconciliation command could report or repair missing and stale rows after role changes, imports, seeders, migrations, or manual database updates.

[...]

PR #21, opened 2026-08-20
Role-based access management in course search

NEOGDAWG:
This PR adds role-based access management to Course Search so users only see courses they have access to.

- Added direct course access filtering using `course_users`.
- Added role-based access filtering using the existing `course_user_role` rows for Program Directors, Department Heads, and faculty-wide access.
- Added an administrator bypass so administrators can search all courses.
- Updated course and program filter options to only include accessible courses.
- Updated Program View so programs only appear when they contain an accessible course.
- Updated program links so inaccessible programs are displayed without a link.
- Applied the same access restrictions to indexed course material-content search.
- Added and updated tests for each access type and related edge cases.
- Added an optional performance seeder and documented the performance-testing results.
- Updated the search and role-access documentation.

Notes:

- Search controls course visibility, while existing permission levels still determine whether a user can view or edit a course.
- Role-based search access relies on the existing materialized `course_user_role` rows.
- Remaining access synchronization edge cases have been documented: See  #20 

Tests:

- Ran the complete Composer test suite: 166 tests passed with 518 assertions.
- Manually verified authenticated material-content searches for administrator and no-access accounts.

Please refer to `docs/search.md`, `docs/role-based-access-management.md`, and the performance-testing report for more details.


laravel/app/Http/Controllers/SearchController.php
```diff
@@ -86,11 +90,21 @@ public function index(Request $request){
...
             ->get();
 
+        // Program page access can come from direct permissions or an elevated role.
+        $programPageAccessIds = $user->programs()
+            ->pluck('programs.program_id')
+            ->merge($user->programsWithElevatedRoleAccess()->pluck('programs.program_id'))
+            ->map(fn ($programId) => (int) $programId)
+            ->unique()
+            ->values()
+            ->all();
```
FabianLCH: I think you can simplify down using the `allPrograms` method defined for `User` objects, which already merges direct-access and role-based-access programs:
```suggestion
        $programPageAccessIds = $user->allPrograms()->pluck('program_id');
```

laravel/database/seeders/SearchPerformanceSeeder.php
FabianLCH: To make the code more readable, I'm thinking we can add a couple of comments here to explain the main strategies you used to deterministically generate courses, programs, topics, etc.

laravel/tests/Feature/SearchTest.php
```diff
@@ -61,6 +64,352 @@ public function test_guest_user_is_redirected_from_search_page()
...
+        $this->assertSearchVisibility($response, [
+            'Directorium Visible Course',
+        ], [
+            'Directorium Hidden Course',
+        ]);
+        $response->assertSee('Courses: 1');
+    }
+
+    public function test_program_role_without_course_role_does_not_grant_search_access()
+    {
```
FabianLCH: Though the test correctly verifies the current behaviour, I feel this should not be the case. A program director should have access to all courses attached to the program(s) they are the director of

laravel/tests/Feature/SearchTest.php
FabianLCH: I think we can add a few tests here:

- all non-admins can see the only names of programs they only have direct access to (no courses) in the "program" view
- program directors of *multiple* programs can see results from all the corresponding programs
- program directors can see filters of programs they have both role-based AND direct access to

FabianLCH (changes requested):
Mostly minor comments to address but the changes are looking good!

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: Updated this to use the existing allPrograms() method. I did not apply the suggestion directly because pluck() returns a Collection, while the existing Blade checks use in_array(), which requires a native array. I added "->all()" to preserve the current expected type.

laravel/database/seeders/SearchPerformanceSeeder.php
NEOGDAWG: added!

laravel/tests/Feature/SearchTest.php
NEOGDAWG: Added all three tests. The direct-program cases exposed that program names and filter options were only being included when the program contained an accessible course, even if the user had direct access to the program itself. I updated the shared program access query to include direct and role-based program access while keeping course results restricted by course access.

FabianLCH (approved):
We're good to go, thanks Kabir!

PR #23, opened 2026-08-27
Kabir/search results export

NEOGDAWG:
This PR adds PDF export functionality to Course Search so users can download and share the results of useful searches.

- Added a Download PDF button to Course Search results.
- Added PDF exports for both Course View and Program View.
- Included the search query, selected filters, matching courses/programs, highlighted snippets, and material-content source details in the exports.
- Exported all matching results rather than only the current pagination page.
- Added descriptive filenames containing the search query, selected view, and date.
- Added focused tests for search exports and existing course export behaviour.
- Updated `docs/search.md` with PDF export details.

Notes:

- Spreadsheet export was considered but removed from the final implementation because PDF better represents the ranked results, grouped programs, and matching snippets shown by Course Search.
- Very broad searches containing thousands of results may need to be narrowed before export because PDF generation has a higher memory cost.

Tests:

- `php artisan test tests/Feature/CourseExportTest.php tests/Feature/SearchTest.php`
- 65 tests passed with 271 assertions.

Please refer to `docs/search.md` for more information about Course Search exports.

### DIFF
--- a/laravel/app/Http/Controllers/CourseController.php
+++ b/laravel/app/Http/Controllers/CourseController.php
@@ -784,10 +784,11 @@ class CourseController extends Controller
             $pdf = PDF::loadView('courses.downloadSummary', compact('course', 'courseLearningOutcomes', 'programsLearningOutcomes', 'unCategorizedProgramsLearningOutcomes', 'programsMappingScales', 'outcomeActivities', 'outcomeAssessments', 'courseStandardOutcomes', 'courseStandardScales', 'standardOutcomeMap', 'assessmentMethodsTotal', 'courseProgramsOutcomeMaps', 'optionalSubcategories'));
             // get the content of the pdf document
             $content = $pdf->output();
+            $pdfPath = 'course-'.$course->course_id.'.pdf';
             // store the pdf document in storage/app/public folder
-            Storage::put('public/course-'.$course->course_id.'.pdf', $content);
+            Storage::disk('public')->put($pdfPath, $content);
             // get the url of the document
-            $url = Storage::url('course-'.$course->course_id.'.pdf');
+            $url = Storage::disk('public')->url($pdfPath);
 
             // return the location of the pdf document on the server
             return $url;
@@ -807,7 +808,7 @@ class CourseController extends Controller
 
     public function deletePDF(Request $request, $course_id)
     {
-        Storage::delete('public/course-'.$course_id.'.pdf');
+        Storage::disk('public')->delete('course-'.$course_id.'.pdf');
     }
 
     // Method for generating data excel in course level
@@ -933,13 +934,13 @@ class CourseController extends Controller
             // generate the spreadsheet
             $writer = new Xlsx($spreadsheet);
             // set the spreadsheets name
-            $spreadsheetName = 'data-summary-'.$course->course_title.'.xlsx';
-            // create absolute filename
-            $storagePath = storage_path('app'.DIRECTORY_SEPARATOR.'public'.DIRECTORY_SEPARATOR.'spreadsheets'.DIRECTORY_SEPARATOR.$spreadsheetName);
+            $spreadsheetName = 'data-summary-course-'.$course->course_id.'.xlsx';
+            $spreadsheetPath = 'spreadsheets/'.$spreadsheetName;
+            Storage::disk('public')->makeDirectory('spreadsheets');
             // save the spreadsheet document
-            $writer->save($storagePath);
+            $writer->save(Storage::disk('public')->path($spreadsheetPath));
             // get the url of the document
-            $url = Storage::url('spreadsheets'.DIRECTORY_SEPARATOR.$spreadsheetName);
+            $url = Storage::disk('public')->url($spreadsheetPath);
 
             // return the location of the spreadsheet document on the server
             return $url;
@@ -954,6 +955,11 @@ class CourseController extends Controller
 
             return -1;
         }
+    }
+
+    public function deleteDataSpreadsheet(Request $request, $course_id)
+    {
+        Storage::disk('public')->delete('spreadsheets/data-summary-course-'.$course_id.'.xlsx');
     }
 
 
--- a/laravel/app/Http/Controllers/CourseController.php
+++ b/laravel/app/Http/Controllers/CourseController.php
@@ -69,7 +69,7 @@ class CourseController extends Controller
     public function __construct()
     {
         $this->middleware(['auth', 'verified']);
-        $this->middleware('course')->only(['show', 'pdf', 'edit', 'submit', 'outcomeDetails']);
+        $this->middleware('course')->only(['show', 'pdf', 'dataSpreadsheet', 'deletePDF', 'deleteDataSpreadsheet', 'edit', 'submit', 'outcomeDetails']);
         $this->roleAssignmentHelper = new RoleAssignmentHelpers();
     }
 
--- a/laravel/app/Http/Controllers/ProgramController.php
+++ b/laravel/app/Http/Controllers/ProgramController.php
@@ -962,12 +962,13 @@ class ProgramController extends Controller
             $content = $pdf->output();
             // set name of pdf
             $pdfName = 'summary-'.$program->program_id.'.pdf';
+            $pdfPath = 'pdfs/'.$pdfName;
             // store the pdf document in storage/app/public folder
-            Storage::put('public'.DIRECTORY_SEPARATOR.'pdfs'.DIRECTORY_SEPARATOR.$pdfName, $content);
+            Storage::disk('public')->put($pdfPath, $content);
             // delete charts
             $this->deleteCharts($program_id, $charts);
             // get the url of the document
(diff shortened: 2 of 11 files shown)


laravel/app/Http/Controllers/SearchController.php
```diff
@@ -207,6 +189,189 @@ public function index(Request $request){
...
+            'properties' => ['nullable', 'array'],
+            'properties.*' => [SearchFilterOptions::propertyValidationRule()],
+            'course_filters_applied' => ['nullable', 'boolean'],
+            'course_codes' => ['nullable', 'array'],
+            'course_codes.*' => ['nullable', 'string', 'max:10'],
+            'course_levels' => ['nullable', 'array'],
+            'course_levels.*' => ['nullable', 'in:100,200,300,400,500,600'],
+            'program_filters_applied' => ['nullable', 'boolean'],
+            'program_ids' => ['nullable', 'array'],
+            'program_ids.*' => ['nullable', 'integer'],
```
FabianLCH: Since most of these validation rules are also used in `index()`, I'm thinking we can extract the shared rules into a class property and extend/modify the rule set as needed. That way, if we make changes to validation in the future we only need to worry about making those changes in one place

laravel/resources/views/search/exports/course-results.blade.php
```diff
@@ -0,0 +1,130 @@
...
+            </tr>
+        </table>
+
+        @forelse($results as $result)
+            <div class="result">
+                <h2>{{ $result->course_code }} {{ $result->course_num }}: {{ $result->course_title }}</h2>
+
+                <div class="programs">
+                    <strong>Programs:</strong>
+                    {{ $result->programs->pluck('program')->implode(', ') ?: 'None' }}
```
FabianLCH: Let's render the programs as an unordered list for cleaner visual presentation!

laravel/resources/views/search/exports/course-results.blade.php
```diff
@@ -0,0 +1,130 @@
...
+                </div>
+
+                <div class="match-counts">
+                    <strong>Found in:</strong>
+                    @if($result->is_course_match) Course Identity: 1; @endif
```
FabianLCH: Maybe we can display something like "Direct Name Match" as normal text or a badge here. I'm thinking that might convey the intended meaning better

laravel/resources/views/search/exports/program-results.blade.php
```diff
@@ -0,0 +1,157 @@
...
+                    <strong>Matching courses:</strong> {{ $programResult->courses->count() }}
+                </div>
+
+                @forelse($programResult->courses as $course)
+                    <div class="course">
+                        <h3>{{ $course->course_code }} {{ $course->course_num }}: {{ $course->course_title }}</h3>
+
+                        <div class="match-counts">
+                            <strong>Found in:</strong>
+                            @if($course->is_course_match) Course Identity: 1; @endif
```
FabianLCH: Similar comment here as in `course-results.blade.php`

laravel/app/Http/Controllers/CourseController.php
```diff
@@ -956,6 +957,11 @@ public function dataSpreadsheet(Request $request, $course_id)
         }
     }
 
+    public function deleteDataSpreadsheet(Request $request, $course_id)
+    {
+        Storage::disk('public')->delete('spreadsheets/data-summary-course-'.$course_id.'.xlsx');
+    }
```
FabianLCH: It doesn't look like this function is being used at the moment. Is this something that is still in the works?

laravel/app/Http/Controllers/ProgramController.php
```diff
@@ -2585,6 +2586,11 @@ public function delSpreadsheet(Request $request, int $programId)
         }
     }
 
+    public function deleteDataSpreadsheet(Request $request, int $programId)
+    {
+        Storage::disk('public')->delete('spreadsheets/data-summary-program-'.$programId.'.xlsx');
+    }
```
FabianLCH: Similarly, this function isn't being used. Is it also a leftover?

FabianLCH (changes requested):
Looking good! Two extra things I'd like to highlight:

- Providing the option for spreadsheet export might still be useful. I can imagine some users wanting to dig deeper on the results, which the spreadsheet format would enable
    - To include all the results in one 'package', we could have spreadsheet tabs split similar to the following example:
        - _Search parameters tab:_ lists search term, filters and overall stats
        - _Search summary tab:_ lists course/program matches and their individual stats
        - _(Individual property matches):_ added conditionally. there would be one for each property that has at least one match
     - Let me know your thoughts on this!
- We should set a hard limit for the maximum number of results to render in detail in the PDF
    - We could set it to a configurable value (via environment variables, defaulting to something like 500 courses) and add a note to the exported PDF that mentions results were truncated due to the limit
    - The search stats should still reflect the full dataset

NEOGDAWG:
> Looking good! Two extra things I'd like to highlight:
> 
> * Providing the option for spreadsheet export might still be useful. I can imagine some users wanting to dig deeper on the results, which the spreadsheet format would enable
>   
>   * To include all the results in one 'package', we could have spreadsheet tabs split similar to the following example:
>     
>     * _Search parameters tab:_ lists search term, filters and overall stats
>     * _Search summary tab:_ lists course/program matches and their individual stats
>     * _(Individual property matches):_ added conditionally. there would be one for each property that has at least one match
>   * Let me know your thoughts on this!
> * We should set a hard limit for the maximum number of results to render in detail in the PDF
>   
>   * We could set it to a configurable value (via environment variables, defaulting to something like 500 courses) and add a note to the exported PDF that mentions results were truncated due to the limit
>   * The search stats should still reflect the full dataset

I agree, I think adding the spreadsheet export could definitely still be useful. I’ll include that alongside the PDF export and also work on adding the configurable hard limit for detailed PDF results!

laravel/app/Http/Controllers/SearchController.php
NEOGDAWG: Done! I extracted the shared search and export filter validation rules into a new class.

laravel/resources/views/search/exports/course-results.blade.php
NEOGDAWG: Done!

laravel/app/Http/Controllers/CourseController.php
NEOGDAWG: These methods are being used through the DELETE routes. After a spreadsheet download starts, the frontend waits 15 seconds and sends a DELETE request to the same URL, which calls the corresponding deleteDataSpreadsheet() method. I renamed the shared JavaScript cleanup function from deletePDF() to deleteGeneratedFile() to make this clearer.

NEOGDAWG:
Hi! I’ve addressed the review feedback, added spreadsheet exports and configurable PDF result limiting (through env), and completed a final cleanup and testing pass. I also added export loading states and ensured PDF and spreadsheet downloads can run simultaneously (this was an edge case that was causing bugs for my browser). The Search test suite is passing, and the export flows have been manually verified. Please let me know if the exported spreadsheet structure and layout is missing anything.

laravel/app/Exports/SearchResultsSpreadsheet.php
```diff
@@ -0,0 +1,294 @@
...
+
+    private function buildCourseSummary(Worksheet $sheet, $results): void
+    {
+        $headings = [
+            'Course Code',
+            'Course Number',
+            'Course Title',
+            'Programs',
+            'Relevance Score',
+            'Course Identity Match',
```
FabianLCH: Maybe we can change this to "Direct Name Match?" for more clarity

laravel/app/Exports/SearchResultsSpreadsheet.php
```diff
@@ -0,0 +1,294 @@
...
+        $this->formatSheet($sheet, count($rows) + 1, count($headings));
+    }
+
+    private function buildProgramSummary(Worksheet $sheet, $programResults): void
+    {
+        $headings = [
+            'Program',
+            'Program Name Match',
+            'Matching Courses',
+            'Course Identity Matches',
```
FabianLCH: Here, let's use "Direct Course Name Matches"

laravel/app/Http/Controllers/SearchController.php
```diff
@@ -197,6 +189,156 @@ public function index(Request $request){
...
+     * Limits the detailed course entries rendered in a PDF without changing the full search data.
+     *
+     * @param array $searchData The complete access-controlled search results and statistics.
+     * @param string $selectedView The selected course or program result view.
+     *
+     * @return array The limited PDF collections and details describing whether they were truncated.
+     */
+    private function limitPdfResults(array $searchData, string $selectedView): array
+    {
+        $limit = max(1, (int) config('search.pdf_result_limit', 500));
```
FabianLCH: You can omit the `max` call here and go with `config('search.pdf_result_limit')` since you already have both a positive value check and a default value in `config/search.php`

FabianLCH (approved):
Just commented on some very minor things, but this PR is good to merge otherwise!

